PR #21, opened 2021-03-29
Add support for multi-file models

### LINKED ISSUE #11: Support multi-file models
The LSP currently only supports a single model file.

Support should be added for using multiple files within a workspace.

srchase:
Closes #11 

Adds support for multi-file models.

When compiling a model, all Smithy model files within the workspace are pulled into the model assembler.  Any files in the `/build` directory are excluded to prevent artifacts from prior builds being included and causing duplication.

With this change, go-to-definition works across multiple files, as does validation of models that span multiple input files.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
@@ -16,10 +16,13 @@
 package software.amazon.smithy.lsp;
 
 import java.io.File;
-
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.commons.io.FileUtils;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
-
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.loader.ModelAssembler;
 import software.amazon.smithy.model.validation.ValidatedResult;
 
 public final class SmithyInterface {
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
```diff
@@ -29,14 +32,31 @@ private SmithyInterface() {
...
       return Either.forLeft(e);
     }
   }
+
+  private static List<File> getModelFiles(File root) {
+    Collection<File> smithyModelFiles = FileUtils.listFiles(root, new String[]{"smithy"}, true);
+    // Filter out any any model files that have been generated during a previous build.
+    return smithyModelFiles.stream().filter(f -> !f.getAbsolutePath().contains(root.getAbsolutePath() + "/build"))
+            .collect(Collectors.toList());
+  }
```
JordonPhillips, line 61: I'm not sure this is the right approach. You could be pulling in a ton of files that aren't necessarily compatible with each other. For instance, projections could have `imports` with conflicting trait or even shape definitions.

What I would expect is that you travel up through directories from the current file's dir up to the workspace root searching for the first `smithy-build.json` and then run the same logic that the gradle plugin uses to load it. There's still potential questions about how to treat projections, but using the source projection will be the best bet. Maybe later some option to switch which projection you're looking at could be added.

JordonPhillips (changes requested):

PR #35, opened 2022-04-15
Determine end of definition location

srchase:
Currently, the source location of a shape is used to set both the start and end positions of a location range when building the map of definition locations used by the definition provider.

This CR improves the definition locations by finding the appropriate end position of a shape. By working from the bottom of a file to the top, the real end of a shape definition can be found by ignoring the traits and comments of the shape beneath it in a file.

This also adds a `getShapeIdFromLocation` method which returns the shapeId of the shape defined at the given location within the model.

With improved definition locations, clients can render more accurate definitions.

In VS Code, the Peek Definition pane currently shows the following definition for a shape:
current

With the improved definition location, the entire shape is highlighted:
updated

When using the "Go to Definition" context option, the cursor still moves to the beginning of the shape. With this CR, when selecting "Go to Definition" when the cursor is already within a shape, the cursor will no longer move. Currently, the cursor moves to the beginning of the shape definition due to the definition location being a zero-length location range that precedes the actual shape definition.

This CR also cleans up the unused `reload` method in SmithyProject.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
@@ -262,6 +262,13 @@ public class SmithyTextDocumentService implements TextDocumentService {
     public CompletableFuture<Either<List<? extends Location>, List<? extends LocationLink>>> definition(
             DefinitionParams params) {
         try {
+            // This attempts to return the definition location that corresponds to a position within a text document.
+            // First, the position is used to find any shapes in the model that are defined at that location. Next,
+            // a token is extracted from the raw text document. The model is walked from the starting shapeId and any
+            // the locations of neighboring shapes that match the token are returned. For example, if the position
+            // is the input of an operation, the token will be the name of the input structure, and the operation will
+            // be walked to return the location of where the input structure is defined. This allows go-to-definition
+            // to jump from the input of the operation, to where the input structure is actually defined.
             List<Location> locations;
             Optional<ShapeId> initialShapeId = project.getShapeIdFromLocation(params.getTextDocument().getUri(),
                     params.getPosition());
@@ -279,7 +286,9 @@ public class SmithyTextDocumentService implements TextDocumentService {
                 // Use location on target, or else default to initial shape.
                 locations = Collections.singletonList(project.getLocations().get(target.orElse(initialShapeId.get())));
             } else {
-                // If cursor location doesn't correspond to definition, return locations of all shapes that match token.
+                // If the definition params do not have a matching shape at that location, return locations of all
+                // shapes that match token by shape name. This makes it possible link the shape name in a line
+                // comment to its definition.
                 locations = project.getLocations().entrySet().stream()
                         .filter(entry -> entry.getKey().getName().equals(found))
                         .map(entry -> entry.getValue())
--- a/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
@@ -206,7 +206,11 @@ public class SmithyTextDocumentService implements TextDocumentService {
                 LspLog.println("Path " + path + " was found in temporary buffer");
                 contents = Arrays.stream(tempContents.split("\n")).collect(Collectors.toList());
             } else {
-                contents = readAll(new File(URI.create(path)));
+                try {
+                    contents = readAll(new File(URI.create(path)));
+                } catch (IllegalArgumentException e) {
+                    contents = readAll(new File(path));
+                }
             }
 
         }
@@ -356,7 +360,11 @@ public class SmithyTextDocumentService implements TextDocumentService {
     }
 
     private File fileFromUri(String uri) {
-        return new File(URI.create(uri));
+        try {
+            return new File(URI.create(uri));
+        } catch (IllegalArgumentException e) {
+            return new File(uri);
+        }
     }
 
     /**
--- a/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
@@ -63,6 +63,10 @@ import software.amazon.smithy.lsp.ext.SmithyProject;
 import software.amazon.smithy.lsp.ext.model.SmithyBuildExtensions;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
+import software.amazon.smithy.model.neighbor.Walker;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.validation.ValidatedResult;
 import software.amazon.smithy.model.validation.ValidationEvent;
 
@@ -254,8 +258,30 @@ public class SmithyTextDocumentService implements TextDocumentService {
     public CompletableFuture<Either<List<? extends Location>, List<? extends LocationLink>>> definition(
             DefinitionParams params) {
         try {
+            List<Location> locations;
+            Optional<ShapeId> initialShapeId = project.getShapeIdFromLocation(params.getTextDocument().getUri(),
+                    params.getPosition());
             String found = findToken(params.getTextDocument().getUri(), params.getPosition());
-            return Utils.completableFuture(Either.forLeft(project.getLocations().getOrDefault(found, noLocations)));
+            if (initialShapeId.isPresent()) {
+                Model model = project.getModel().unwrap();
+                Shape initialShape = model.getShape(initialShapeId.get()).get();
+                // Find first neighbor (non-member) with name that matches token.
+                Walker shapeWalker = new Walker(NeighborProviderIndex.of(model).getProvider());
+                Optional<ShapeId> target = shapeWalker.walkShapes(initialShape).stream()
+                        .filter(shape -> !shape.isMemberShape())
+                        .map(shape -> shape.getId())
+                        .filter(shape -> shape.getName().equals(found))
+                        .findFirst();
+                // Use location on target, or else default to initial shape.
+                locations = Collections.singletonList(project.getLocations().get(target.orElse(initialShapeId.get())));
+            } else {
+                // If cursor location doesn't correspond to definition, return locations of all shapes that match token.
+                locations = project.getLocations().entrySet().stream()
+                        .filter(entry -> entry.getKey().getName().equals(found))
+                        .map(entry -> entry.getValue())
+                        .collect(Collectors.toList());
+            }
+            return Utils.completableFuture(Either.forLeft(locations));
         } catch (Exception e) {
             // TODO: handle exception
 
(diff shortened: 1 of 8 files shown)


src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -161,35 +160,97 @@ public File getRoot() {
...
-                : !fileName.startsWith("file:") ? "file:" + fileName
-                : fileName;
+        List<String> modelFiles = model.shapes()
+                .map(shape -> shape.getSourceLocation().getFilename())
+                .distinct()
+                .collect(Collectors.toList());
+        for (String modelFile : modelFiles) {
+            Path modelPath = Paths.get(modelFile);
+            List<String> lines = getFileLines(modelPath);
+            LspLog.println(lines);
```
kstich, line 170: Seems verbose to log every line of every file.

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -161,35 +160,97 @@ public File getRoot() {
...
+            int endMarker = 0;
+            try {
```
kstich, line 172: Isn't this just `lines.size()`?

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -161,35 +160,97 @@ public File getRoot() {
...
+            LspLog.println("Path " + path + " was found in temporary buffer");
+        }
+        return Collections.emptyList();
+    }
+
+    private static void addLocationToMap(Shape shape, Map<String, List<Location>> locations, Location location) {
+        String shapeName = shape.getId().getName();
+        // Members get the same shapeName as their parent structure
+        // so we ignore them, to avoid producing a location per-member
+        // TODO: index members somehow as well?
```
kstich, line 235: This seems useful, since `apply` syntax allows for targeting a member externally. Maybe the `location` map should use the `ShapeId` as the key instead of the name?

src/test/java/software/amazon/smithy/lsp/ext/SmithyProjectTest.java
```diff
@@ -71,4 +75,29 @@ public void respectingEmptyConfig() throws Exception {
...
+    public void definitionLocations() throws Exception {
+        Path baseDir = Paths.get(SmithyProjectTest.class.getResource("models").toURI());
+        Path modelMain = baseDir.resolve("main.smithy");
+        Path modelTest = baseDir.resolve("test.smithy");
+        List<Path> modelFiles = ImmutableList.of(modelMain, modelTest);
+
+        try (Harness hs = Harness.create(SmithyBuildExtensions.builder().build(), modelFiles)) {
+            Map<String, List<Location>> locations = hs.getProject().getLocations();
+
+            correctLocation(locations, "SingleLine", 4, 5);
```
kstich, line 88: Seems a little strange that the starts for this are the line before, worth seeing if that's necessary to work properly.

src/test/resources/software/amazon/smithy/lsp/ext/models/main.smithy
```diff
@@ -0,0 +1,44 @@
+$version: "1.0"
```
kstich, line 1: Might be worth throwing some oddly formatted definitions in here just to make sure they work as well - indented, same line closing braces, etc.

kstich (changes requested):

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
srchase: When finding a token in a line, SmithyTextDocumentService only returns a potential shape name. We could use the shape name to try to find the appropriate `ShapeId` and then change over the location map to use the `ShapeId` as the key.

srchase: When finding a token in a line, SmithyTextDocumentService only returns a potential shape name. We could use the shape name to try to find the appropriate `ShapeId` and then change over the location map to use the `ShapeId` as the key.

src/test/java/software/amazon/smithy/lsp/ext/SmithyProjectTest.java
srchase: Positions for LSP are zero-based, so this is currently setting the end position to the start of the next line.

I'll add logic to properly determine the line length so we can determine the actual end of a shape on a given line.

src/test/resources/software/amazon/smithy/lsp/ext/models/main.smithy
srchase: Added in 7aa59d0

Baccata:
> When using the "Go to Definition" context option, the cursor still moves to the beginning of the shape. With this CR, when selecting "Go to Definition" when the cursor is already within a shape, the cursor will no longer move. Currently, the cursor moves to the beginning of the shape definition due to the definition location being a zero-length location range that precedes the actual shape definition.

I've not read the code so bear with me, but could use these more precise locations to actually determine the actual shape the cursor is hovering over ? `Go to Definition` suffers from not having this information, as it does a naive extraction of the underlying string, then gets you the first shape the name of which matches that string. If we had a `Location => Shape` method, when the cursor is on a member shape, we could jump to the actual targeted shape. There's also a number of additional LSP endpoints that could be implemented by virtue of having this information.

srchase:
> > When using the "Go to Definition" context option, the cursor still moves to the beginning of the shape. With this CR, when selecting "Go to Definition" when the cursor is already within a shape, the cursor will no longer move. Currently, the cursor moves to the beginning of the shape definition due to the definition location being a zero-length location range that precedes the actual shape definition.
> 
> I've not read the code so bear with me, but could use these more precise locations to actually determine the actual shape the cursor is hovering over ? `Go to Definition` suffers from not having this information, as it does a naive extraction of the underlying string, then gets you the first shape the name of which matches that string. If we had a `Location => Shape` method, when the cursor is on a member shape, we could jump to the actual targeted shape. There's also a number of additional LSP endpoints that could be implemented by virtue of having this information.

Yes, the member's location could be used to determine its actual definition which could then allow for jumping to its target. One nice side-effect of the current naive extraction is that tokens can be pulled out of line docs and you can still jump to the definition.
Which other LSP endpoints did you have in mind for being able to use a `Location => Shape` method?

Baccata:
> Which other LSP endpoints did you have in mind for being able to use a Location => Shape method? 

I was thinking things along the lines of : 

* [find references](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_references)
* rename symbol

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -232,4 +359,24 @@ private static List<File> downloadExternalDependencies(SmithyBuildExtensions ext
     private static List<File> onlyExistingFiles(Collection<File> files) {
         return files.stream().filter(File::isFile).collect(Collectors.toList());
     }
+
+    // TODO: Remove this Class once the change in https://github.com/awslabs/smithy/pull/1192 is available.
```
JordonPhillips, line 363: This is available now

JordonPhillips (changes requested):

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
srchase: This got merged after 1.21.0. It can get picked up once 1.22.0 is released.

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -159,37 +161,162 @@ public File getRoot() {
...
+        } catch (IOException e) {
+            LspLog.println("File " + file + " could not be loaded.");
+        }
+        return Collections.emptyList();
+    }
+
+    private static String getUri(String fileName) {
+        return Utils.isJarFile(fileName)
+                ? Utils.toSmithyJarFile(fileName)
+                : !fileName.startsWith("file:") ? "file:" + fileName
```
kstich, line 316: Consider breaking this nested ternary out for readability.

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -159,37 +161,162 @@ public File getRoot() {
...
+                // Since the position is in each of the overlapping shapes, return the smallest range.
+                .sorted(rangeSize)
+                .collect(Collectors.toList());
+        if (matchingShapes.isEmpty()) {
+            return Optional.empty();
+        }
+        return Optional.ofNullable(matchingShapes.get(0).getKey());
+    }
+
+    private Predicate<Map.Entry<ShapeId, Location>> filterByLocation(Position position) {
```
kstich, line 262: This seems like it should be a `Predicate<Location>` instead, since it doesn't use the key at all.

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -159,37 +161,162 @@ public File getRoot() {
...
-                    locList.add(location);
-                    locations.put(shapeName, locList);
+                    endPosition = getEndPosition(endMarker, lines);
                 }
-            }
-        });
 
+                // Find the end of a member's location by first trimming trailing commas, empty lines and closing
+                // structure braces.
+                if (shape.getType() == ShapeType.MEMBER) {
```
kstich, line 196: Is it reasonable to break these branches up in to their own methods for readability?

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -254,8 +262,30 @@ private String findToken(String path, Position p) throws IOException {
...
+                Walker shapeWalker = new Walker(NeighborProviderIndex.of(model).getProvider());
+                Optional<ShapeId> target = shapeWalker.walkShapes(initialShape).stream()
+                        .filter(shape -> !shape.isMemberShape())
+                        .map(shape -> shape.getId())
+                        .filter(shape -> shape.getName().equals(found))
+                        .findFirst();
+                // Use location on target, or else default to initial shape.
+                locations = Collections.singletonList(project.getLocations().get(target.orElse(initialShapeId.get())));
+            } else {
+                // If cursor location doesn't correspond to definition, return locations of all shapes that match token.
```
kstich, line 282: Not sure what this comment means.

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -254,8 +262,30 @@ private String findToken(String path, Position p) throws IOException {
     public CompletableFuture<Either<List<? extends Location>, List<? extends LocationLink>>> definition(
             DefinitionParams params) {
         try {
+            List<Location> locations;
```
kstich, line 265: It would be nice to have a description of what this is doing behaviorally.

srchase: Added with an example.

kstich (approved):

PR #40, opened 2022-05-12
Fix location for trait applied in other file

srchase:
This PR fixes #39.

When determining the end location of a shape definition, traits that do not have a source location (i.e `SourceLocation.NONE`), or have a source location in another model file (when using an `apply` statement) should be ignored when advancing the end marker used to determine the end location of the next shape to be parsed.

A `TODO` has been added to address that `apply` statements that precede the definition of a shape within the same model file will result in invalid locations. A subsequent PR should address this by ensuring that a trait's source location should only be used to advance the end marker if that location is not higher in a file than the starting location of the next shape to be parsed.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #43, opened 2022-05-12
Fix definition of trait on member

### LINKED ISSUE #42: Following annotation on shape's member does not work
TLDR: CMD + click on an annotation on a structure's member will not be followed

See before:

https://user-images.githubusercontent.com/5230460/168094104-5821dac0-40e2-4228-8aaa-e1dad6f40002.mov

See after: (this is based on a build at commit: e82fbffb1a1de7bd1bba0e3dd378c5bba46e9e76. this is not the commit that introduced the issue, it's just a reference that you can use if you want to reproduce.)

https://user-images.githubusercontent.com/5230460/168094213-5822ceb4-8241-4020-916f-06d2adc6a55f.mov

In the first video, we can see that CMD + click on an annotation on a structure's member will be followed (`@required` for example).

In the second video, we see that clicking on such an annotation does not work anymore. Clicking on annotation on top level shapes will work fine.

I don't have the bandwidth to troubleshoot the issue right now. I will come back to it later.

srchase:
This PR fixes #42.

This fixes returning the correct definition location for a trait applied to a member. It also adds a test for the definition location of a trait at the top level.

Previously, only the non-member shapes were walked to find the target of a specific location within a model. This adds logic to also check the traits applied to members within the shape.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
@@ -283,6 +283,15 @@ public class SmithyTextDocumentService implements TextDocumentService {
                         .map(Shape::getId)
                         .filter(shape -> shape.getName().equals(found))
                         .findFirst();
+                // If none of the neighbors match the token, check traits applied to members of the shape.
+                if (!target.isPresent()) {
+                    target = shapeWalker.walkShapes(initialShape).stream()
+                            .filter(shape -> shape.isMemberShape())
+                            .flatMap(shape -> shape.getAllTraits().values().stream())
+                            .map(trait -> trait.toShapeId())
+                            .filter(shapeId -> shapeId.getName().equals(found))
+                            .findFirst();
+                }
                 // Use location on target, or else default to initial shape.
                 locations = Collections.singletonList(project.getLocations().get(target.orElse(initialShapeId.get())));
             } else {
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -283,6 +283,15 @@ public CompletableFuture<Either<List<? extends Location>, List<? extends Locatio
                         .map(Shape::getId)
                         .filter(shape -> shape.getName().equals(found))
                         .findFirst();
+                // If none of the neighbors match the token, check traits applied to members of the shape.
+                if (!target.isPresent()) {
+                    target = shapeWalker.walkShapes(initialShape).stream()
```
kstich, line 288: Does this need to be distinct from the previous `walkShapes` call? They seem to be doing similar work with a separate `map` call if a shape is a member or not, which means they could be reconciled in to one.

daddykotex:
Happy to report that both #43  and #40 address all of the issues I have found with the latest change. When they're merged, we'll cut a release for via our fork, and that should be pulled by the [Smithy extension](https://marketplace.visualstudio.com/items?itemName=disneystreaming.smithy)

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
srchase: Good call. Combined in [f3cda43](https://github.com/awslabs/smithy-language-server/pull/43/commits/f3cda43be9595b7363109d6bc6b0f04e2051f5d5).

kstich (approved):

PR #44, opened 2022-05-24
Do not load dependencies twice

srchase:
Currently, when a trait is used in a model that requires loading a dependency jar, an error like the following may occur:
```
java.lang.ClassCastException: class software.amazon.smithy.test.traits.TestTrait cannot be cast to class software.amazon.smithy.test.traits.TestTrait (software.amazon.smithy.test.traits.TestTrait is in unnamed module of loader java.net.URLClassLoader @62727399; software.amazon.smithy.test.traits.TestTrait is in unnamed module of loader java.net.URLClassLoader @7c098bb3)
```

This error occurs only on traits which use a Provider class like the following:
```
    public static final class Provider extends AbstractTrait.Provider {
        public Provider() {
            super(ID);
        }

        @Override
        public Trait createTrait(ShapeId target, Node value) {
            TestTrait result = new NodeMapper().deserialize(value, TestTrait.class);
            result.setNodeCache(value);
            return result;
        }
    }
```

For traits like the above, the resulting `NodeMappter().deserialize()` call requires a cast to the resulting trait type. In the Language Server, on reload, the dependency jar is loaded multiple times which causes the cast to fail when it occurs between different instances of the loaded dependency.

This fix updates the SmithyInterface class to prevent re-loading dependency jars multiple times and instead relies on loading the dependencies into the upstream model before the model is assembled.


kstich (approved):

JordonPhillips (approved):

daddykotex:
thanks for the fix, do you want me to test it out with the VS Code extension?

As I said to @srchase, I'm not sure why the code is the way it is, especially because it does not really reflect my understanding from the comment above. So I'd be happy to see that duplication gone. Even better if that fixes a bug!

srchase:
@daddykotex I took testing pass with the VS Code extension already.

PR #45, opened 2022-06-02
Improve auto-completions for traits

srchase:
This PR improves the auto-completions for traits.

Previously, when starting type `@http`, any traits or shapes prefixed with `http` would be returned as possible completions.

Following this change, when starting to type `@http` to add one of the @http* traits to an operation, the following completions will be returned: `http(method: "", uri: "")`, `http()` and `httpChecksumRequired`.

The first, `http(method: "", uri: "")`, includes the required members of the [@http trait](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#http-trait) and sets the target strings to empty strings. The `code` member is not required and is omitted.

The second, `http()`, is included as a default, with none of the required or optional members populated.

The third, `httpChecksumRequired`, does not include the set of parentheses since [this](https://awslabs.github.io/smithy/1.0/spec/core/behavior-traits.html#httpchecksumrequired-trait) annotation trait does not specify any members.

None of the other @http* traits are suggested (`httpError`, `httpHeader`, etc.) since they cannot be applied to an operation.


src/main/java/software/amazon/smithy/lsp/ext/Completions.java
```diff
@@ -41,24 +62,56 @@ private Completions() {
...
+        // If the token is part of a trait statement, filter the set to trait shapes which can be applied to the shape
+        // that the trait targets.
+        if (isTrait) {
+            shapeIdSet = model.shapes()
+                    .filter(shape -> shape.hasTrait(ShapeId.from("smithy.api#trait")))
+                    .filter(shape -> {
+                        if (!target.isPresent()) {
+                            return true;
+                        }
+                        return shape.getTrait(TraitDefinition.class).get().getSelector().shapes(model)
```
kstich, line 84: Can use `expectTrait` since it was filtered before.

src/main/java/software/amazon/smithy/lsp/ext/Completions.java
```diff
@@ -41,24 +62,56 @@ private Completions() {
...
+                                .anyMatch(matchingShape -> matchingShape.getId().equals(target.get()));
+                    })
+                    .map(shape -> shape.getId())
+                    .collect(Collectors.toSet());
+        } else {
+            // Otherwise, use all shapes in the as potential completions.
```
kstich, line 90: s/the as/the model as

src/main/java/software/amazon/smithy/lsp/ext/Completions.java
```diff
@@ -41,24 +62,56 @@ private Completions() {
...
+    public static List<SmithyCompletionItem> find(Model model, String token, boolean isTrait,
+                                                  Optional<ShapeId> target) {
         Map<String, SmithyCompletionItem> comps = new HashMap<>();
         String lcase = token.toLowerCase();
 
+        Set<ShapeId> shapeIdSet;
+        // If the token is part of a trait statement, filter the set to trait shapes which can be applied to the shape
+        // that the trait targets.
+        if (isTrait) {
+            shapeIdSet = model.shapes()
```
kstich, line 78: Can this be its own method?

src/main/java/software/amazon/smithy/lsp/ext/Completions.java
```diff
@@ -104,4 +157,121 @@ private static CompletionItem createCompletion(String s, CompletionItemKind kind
...
+        }
+
+        @Override
+        protected String getDefault(Shape shape) {
+            return "";
+        }
+
+        @Override
+        public String bigDecimalShape(BigDecimalShape shape) {
+            return "\"\"";
```
kstich, line 207: Maybe this is a completion thing, that might warrant a comment, but why an empty string for these numbers/timestamp that'd be a number value and not a string value in the IDL?

kstich (changes requested):

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -161,27 +162,132 @@ private MessageParams msg(final MessageType sev, final String cont) {
...
+
+            if (c == ')' && !quote) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    // Work backwards from current position to determine if position is part of a trait.
+    private boolean isTrait(String documentUri, Position position) throws IOException {
```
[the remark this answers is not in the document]
srchase: Yes, this would return `false` if the position is within the trait's value, but this was only intended to return `true` for a position within the trait's shapeId. My assumption there was that if a model author has already added an opening parenthesis and started to write the trait's value, they wouldn't really need completions. Additionally, the above token parser only detects potential shapeIds for now.

src/main/java/software/amazon/smithy/lsp/ext/Completions.java
srchase: For this visitor, I set any shapes that could be represented as strings, per this trait node values [table](https://awslabs.github.io/smithy/1.0/spec/core/model.html#trait-node-values), as `""`.

Is that table out of date?

Since those are likely to be set as a number value, maybe they should use the default case of an empty string like `byte`, `short`, `integer` and `long`.

src/main/java/software/amazon/smithy/lsp/ext/Completions.java
srchase: Addressed in 43489272c27970304970f2d2e616c95c47cc1c00.

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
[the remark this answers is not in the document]
kstich: I think supplying trait property completions would be nice to have, given [the `aws.apigateway#integration` trait's 15 fields of which only 3 are required](https://awslabs.github.io/smithy/1.0/spec/aws/amazon-apigateway.html#aws-apigateway-integration-trait).

Changing the name for now is probably sufficient though.

src/main/java/software/amazon/smithy/lsp/ext/Completions.java
kstich: At least `float` and `double` should default, since the only valid string values are the edge cases. I think this should use number values for the arbitrary precision as well, since our tooling handles them properly.

I could see leaving `timestamp` as a string value, since it'll be more readable in the model than a Unix epoch.

srchase: Updated `float`, `double`, `bigDecimal` and `bigInteger` to default in 25f2d8e2b1d21b658d6a9132f0511fbf61fb1aa0.

kstich (approved):

PR #49, opened 2022-06-08
Fix use block parsing in preamble

srchase:
This PR fixes the parsing of use statements in the preamble of a model file.

Currently, if any blank lines appear with in the use statement block, parsing of the preamble will only detect the use statements that precede the first blank line.

This issue results in duplicate use statements being added when a trait completion is used.

Take this example model:
```
$version: "1.0"

namespace ns.preamble

use ns.foo#FooTrait

use ns.bar#BarTrait

@FooTrait
string MyString 
```

If `@Bar` was typed to target `MyString`, with the client opting to use the `@BarTrait` completion, it would result in a duplicate `use` statement being added:
```
$version: "1.0"

namespace ns.preamble

use ns.foo#FooTrait
use ns.foo#BarTrait

use ns.bar#BarTrait

@FooTrait
string MyString 
```

This fix corrects the parsing so that the `ns.bar#BarTrait` use statement is correctly identified as already in the model file and prevents it from being added a second time.

### DIFF
--- a/src/test/java/software/amazon/smithy/lsp/ext/DocumentTest.java
+++ b/src/test/java/software/amazon/smithy/lsp/ext/DocumentTest.java
@@ -59,6 +59,8 @@ public class DocumentTest {
         DocumentPreamble preamble = Document.detectPreamble(lines);
 
         assertEquals("ns.preamble", preamble.getCurrentNamespace().get());
+        assertEquals(new Position(2, 0), preamble.getUseBlockRange().getStart());
+        assertEquals(new Position(2, 14), preamble.getUseBlockRange().getEnd());
         assertFalse(preamble.isBlankSeparated());
     }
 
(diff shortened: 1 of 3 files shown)


src/test/java/software/amazon/smithy/lsp/ext/DocumentTest.java
```diff
@@ -0,0 +1,77 @@
...
+                "$version: \"1.0\"",
+                "namespace ns.preamble",
+                "use ns.foo#Foo",
+                "@Foo",
+                "string MyString"
+        );
+        DocumentPreamble preamble = Document.detectPreamble(lines);
+
+        assertEquals("ns.preamble", preamble.getCurrentNamespace().get());
+        assertFalse(preamble.isBlankSeparated());
```
kstich, line 62: Should validate the `getUseBlockRange` here.

srchase: Fixed in 0dc05f0.

kstich (approved):

PR #54, opened 2022-06-15
Create temp dir for standalone model file

srchase:
This PR updates the how the language server is initialized to handle editing a standalone smithy model file. When a client is editing a single model file, a workspace may not have been created yet. This PR checks if the workspace is present in the server initialization params and creates a temporary directory to use in its absence.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
@@ -15,6 +15,7 @@
 
 package software.amazon.smithy.lsp;
 
+import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -64,7 +65,6 @@ public class SmithyLanguageServer implements LanguageServer, LanguageClientAware
 
   @Override
   public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
-    LspLog.println(params.getWorkspaceFolders());
     if (params.getRootUri() != null) {
       try {
         workspaceRoot = new File(new URI(params.getRootUri()));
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -77,6 +77,18 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
       LspLog.println("Workspace root was null");
     }
 
+    if (params.getWorkspaceFolders() == null) {
+      try {
+        File temp = Files.createTempDirectory("tmp-smithy-language-server-workspace").toFile();
+        System.out.println("Created temporary workspace root: " + temp);
+        temp.deleteOnExit();
+        WorkspaceFolder workspaceFolder = new WorkspaceFolder(temp.toURI().toString());
+        params.setWorkspaceFolders(ImmutableList.of(workspaceFolder));
```
DavidOgunsAWS, line 86: An editor could add/remove workspace folders at any time and I'm not sure if lsp4j will defensively copy this list. If it doesn't and tries to modify the same list that'll be a sure bug (exception thrown) later. If it's too hard to verify should be happening here I'd be more comfortable with a mutable list instead.

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -77,6 +77,18 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
       LspLog.println("Workspace root was null");
     }
 
+    if (params.getWorkspaceFolders() == null) {
+      try {
+        File temp = Files.createTempDirectory("tmp-smithy-language-server-workspace").toFile();
+        System.out.println("Created temporary workspace root: " + temp);
```
DavidOgunsAWS, line 83: Should this output go somewhere other than stdout directly?

DavidOgunsAWS (commented):
(ignore, moved comment)

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
srchase: Switched to `ListUtils.of` in https://github.com/awslabs/smithy-language-server/pull/54/commits/033ddbafabccfaea4a295f273e9d0e89811b9c77.

DavidOgunsAWS (approved):

Issue #68, opened 2022-09-14
Issue with collectLocations


daddykotex:
So we've hit a bug in some part of the language-server. I found what's triggering it and I have a small reproduction. When the bug happens, the `project` is not loaded and the rest of the IDE experience is poorly degraded.

The cause is the following NPE:

```
[11:47:51] java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.Location.getRange()" because "<local6>" is null
	at software.amazon.smithy.lsp.ext.SmithyProject.collectMemberLocations(SmithyProject.java:274)
	at software.amazon.smithy.lsp.ext.SmithyProject.collectLocations(SmithyProject.java:239)
	at software.amazon.smithy.lsp.ext.SmithyProject.lambda$new$0(SmithyProject.java:70)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at software.amazon.smithy.lsp.ext.SmithyProject.<init>(SmithyProject.java:70)
	at software.amazon.smithy.lsp.ext.SmithyProject.load(SmithyProject.java:174)
	at software.amazon.smithy.lsp.ext.SmithyProject.load(SmithyProject.java:145)
```

This is rather explicit: we pull something out of a map, it can be null if the thing is not in the map, but we don't check for that, so it blows up.

The quick fix is to do a null check. The right fix is to find why this happens. So I went on a hunt.

Given the following:

```
.
├── core
│   └── common.smithy
└── test
    └── test.smithy
```

With the following content:

```
> cat core/common.smithy 
$version: "2"

namespace core

boolean IsTestInput

@mixin
structure HasIsTestParam {
  isTest: IsTestInput = false
}
```

```
> cat test/test.smithy 
$version: "2"

namespace test

use core#HasIsTestParam

service AService {
  version: "1",
  operations: [SomeOp]
}
operation SomeOp {
  input: SomeOpInput
}

structure SomeOpInput with [HasIsTestParam] {
  @required
  body: String
}

apply SomeOpInput$isTest @documentation("Some documentation")
```

This specific workspace will blow up the LSP w/ the error above. 

Everything is fine if I comment the following line: `apply SomeOpInput$isTest @documentation("Some doc")`

Also, everything is fine if I pull the shapes from `core` into `test` (placing the shapes in the same file / same namespace).

daddykotex:
While debugging, I tweaked the implementation of LspLog a little bit with two improvements:

[...]

PR #69, opened 2022-09-14
LspLog: one fix, one improvement

daddykotex:
1. When the file does not exist the `message` is added to a buffer to eventually be written to the file. When this happens, the `message` is timestamped twice.

2. When println receives an exception, it would be great if the stack trace was printed rather than only the `getMessage`.

This commit address these two things

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/ext/LspLog.java
+++ b/src/main/java/software/amazon/smithy/lsp/ext/LspLog.java
@@ -110,15 +110,13 @@ public final class LspLog {
     private static String processMessage(Object message) {
         if (message == null) {
             return "null";
+        } else if (message instanceof Throwable) {
+            StringWriter sw = new StringWriter();
+            PrintWriter pw = new PrintWriter(sw);
+            ((Throwable) message).printStackTrace(pw);
+            return sw.toString();
         } else {
-            if (message instanceof Throwable) {
-                StringWriter sw = new StringWriter();
-                PrintWriter pw = new PrintWriter(sw);
-                ((Throwable) message).printStackTrace(pw);
-                return sw.toString();
-            } else {
-                return message.toString();
-            }
+            return message.toString();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


daddykotex:
Related to the investigation of: https://github.com/awslabs/smithy-language-server/issues/68

src/main/java/software/amazon/smithy/lsp/ext/LspLog.java
```diff
@@ -90,18 +92,33 @@ private static String currentTime() {
...
-                    buffer.ifPresent(buf -> buf.add(timestamped));
+                    buffer.ifPresent(buf -> buf.add(sanitizedMessage));
                 }
             }
         } catch (Exception e) {
 
         }
     }
+
+    private static String processMessage(Object message) {
```
milesziemer, line 110:
```suggestion
    private static String getStringifiedMessage(Object message) {
```
What do you think about using slightly clearer naming here?

src/main/java/software/amazon/smithy/lsp/ext/LspLog.java
```diff
@@ -90,18 +92,33 @@ private static String currentTime() {
...
+        } else {
+            if (message instanceof Throwable) {
+                StringWriter sw = new StringWriter();
+                PrintWriter pw = new PrintWriter(sw);
+                ((Throwable) message).printStackTrace(pw);
+                return sw.toString();
+            } else {
+                return message.toString();
+            }
+        }
```
milesziemer, line 122:
```suggestion
        if (message == null) {
            return "null";
        } else if (message instanceof Throwable) {
            StringWriter sw = new StringWriter();
            PrintWriter pw = new PrintWriter(sw);
            ((Throwable) message).printStackTrace(pw);
            return sw.toString();
        } else {
            return message.toString();
        }
```
I could be missing something, but I believe this would work the same.

daddykotex: absolutely

src/main/java/software/amazon/smithy/lsp/ext/LspLog.java
daddykotex: Sure no problem

daddykotex:
feel free to squash

milesziemer (approved):

PR #70, opened 2022-09-19
Ignore applys

milesziemer:
*Issue #, if available:*
#68 
*Description of changes:*
`apply` statements are now ignored by the language server when collecting definition locations

As mentioned in the issue, the language server was breaking when an apply statement targeted a member mixed in from another namespace. This brought up various issues about how apply statements were being handled, and more generally about how we were collecting shape definition locations. Apply statements could trick the old `collectLocations` method into thinking shapes began or ended in arbitrary file locations, or flat out make it crash. At this time, `apply` statement definition locations don't serve a purpose to the language server, and should be safely ignorable without compromising functionality.

Additional changes:
- Shape definition location collection was moved into its own class behind an interface that we can use to make it easier to add new ways of collecting locations, or new kinds of locations (other than definitions)
- A new test case was added specifically for apply statement behavior, ensuring they are fully ignored
- Various methods were cleaned up, simplified or broken down into smaller methods


src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
```diff
@@ -0,0 +1,367 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * Creates a cache of {@link ModelFile} and uses it to collect the locations of container
+ * shapes in all files, then collects their members.
+ */
+final class FileCachingCollector implements ShapeLocationCollector {
```
milesziemer, line 33: Not obsessed with this name, I'm open to other suggestions

milesziemer, line 33: Not obsessed with this name, other suggestions are welcome

milesziemer: Leaving this for now

Issue #67, opened 2022-08-30
Set allowUnknownTraits to avoid completely breaking if a trait has no definition


milesziemer:
It might be worth more deeply investigating failure behavior for other situations as well to see if there are other things that could cause the language server to completely break.

milesziemer:
It looks like breaking is only occurring when hovering. This is likely due to calling unwrap in `hover()`: https://github.com/awslabs/smithy-language-server/blob/1b3e58dcd22a8432c0b7dbf310f36a7b2d0b1730/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java#L431 We call unwrap in `definition()` too, but it's inside of the try-catch so if the model isn't valid we just send an empty response. 

This issue suggests setting `allowUnknownTraits`, but further testing needs to be done to make sure that this won't disable diagnostics the language server may send about a trait being undefined. We should however make sure that definition and hover still work for valid shapes, and only send empty responses for unknown shapes.

Issue #73, opened 2022-09-27
Add support for packages that aren't in Maven


milesziemer:
This is related to #67 , but would be a more comprehensive fix. If your model uses some packages that aren't available in Maven, builds can be successful but the language server will not be able to resolve the dependencies. One idea on how we might be able to do this is if these dependencies are jars somewhere locally, we could provide these paths to the language server directly.

PR #74, opened 2022-09-28
More gracefully handle unknown traits

milesziemer:
*Issue #, if available:* #67
*Description of changes:*
This commit allows hover and jump to definition to continue functioning
when there are unknown traits in the model, and stops crashes when hovering
when there are unknown traits in the model. Additionally, when hovering
over shapes that use an unknown trait, an error message will be displayed in
the hover content. Tests were also added to verify that:
- A model with unknown traits can be loaded
- Diagnostics are still sent for unknown traits
- Definition for valid shapes still works when there are unknown traits
- Hover for valid shapes still works when there are unknown traits
- Error messages are displayed in hover content for invalid shapes

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
@@ -54,13 +54,20 @@ public final class SmithyInterface {
         // Contrary to model assemblers, model builders do not complain about the
         // duplication of shapes. Shapes will simply overwrite each other in a "last
         // write wins" kind of way.
-
         URLClassLoader urlClassLoader = new URLClassLoader(urlArray);
-        Model upstreamModel = Model.assembler().discoverModels(urlClassLoader).assemble().unwrap();
+        Model upstreamModel = Model.assembler()
+                .putProperty(ModelAssembler.ALLOW_UNKNOWN_TRAITS, true)
+                .discoverModels(urlClassLoader)
+                .assemble()
+                .unwrap();
         builder.addShapes(upstreamModel);
       }
 
-      ModelAssembler assembler = Model.assembler().addModel(builder.build());
+      ModelAssembler assembler = Model.assembler()
+              // We don't want the model to be broken when there are unknown traits,
+              // because that will essentially disable language server features.
+              .putProperty(ModelAssembler.ALLOW_UNKNOWN_TRAITS, true)
+              .addModel(builder.build());
 
       for (File file : files) {
         assembler = assembler.addImport(file.getAbsolutePath());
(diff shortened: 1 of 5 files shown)


src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
```diff
@@ -54,13 +54,17 @@ public static Either<Exception, ValidatedResult<Model>> readModel(Collection<Fil
         // Contrary to model assemblers, model builders do not complain about the
         // duplication of shapes. Shapes will simply overwrite each other in a "last
         // write wins" kind of way.
-
+        // TODO It's not clear if we need ALLOW_UNKNOWN_TRAITS here
```
milesziemer, line 57: Leaving a comment here to get thoughts on this

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -477,7 +476,25 @@ private Optional<Shape> getTargetShape(Shape initialShape, String token, Model m
...
+            contents.append("Can't display shape ").append("`").append(shape.getId().toString()).append("`:");
+            for (ValidationEvent event : validationEvents) {
+                contents.append(System.lineSeparator()).append(event.getMessage());
+            }
+            if (validationEvents.isEmpty()) {
+                contents.append(System.lineSeparator()).append(e);
+            }
+            return contents.toString();
+        }
+    }
```
milesziemer, line 495: Not sure if this is the best way to go, but my thinking was we only wanted to display the error content if absolutely necessary. There may be a better way to do this that I can't think of at the moment.

src/test/java/software/amazon/smithy/lsp/SmithyTextDocumentServiceTest.java
```diff
@@ -149,6 +150,83 @@ public void sendingDiagnosticsToTheClient() throws Exception {
...
+        try (Harness hs = Harness.create(SmithyBuildExtensions.builder().build(), ListUtils.of(modelFilePath))) {
+            SmithyTextDocumentService tds = new SmithyTextDocumentService(Optional.empty(), hs.getTempFolder());
+            StubClient client = new StubClient();
+            tds.createProject(hs.getConfig(), hs.getRoot());
+            tds.setClient(client);
+
+            TextDocumentIdentifier tdi = new TextDocumentIdentifier(hs.file(modelFilename).toString());
+            Hover hover = tds.hover(hoverParams(tdi, 10, 13)).get();
+            MarkupContent hoverContent = hover.getContents().getRight();
+            assertTrue(hoverContent.getValue().startsWith("Can't display shape"));
```
milesziemer, line 226: I don't love the duplication of "Can't display shape"

src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
srchase: I think allowing unknown traits here makes sense, especially given #73 since we know there's a need to allow pulling in other dependencies.

srchase (approved):

srchase (approved):

PR #78, opened 2022-11-17
Load validator & trait factory

daddykotex:
This [diff](https://github.com/awslabs/smithy-language-server/pull/44/files#diff-1860f6e567537a9aa1979a5c2084b684cefeeb469c495ca4bff0700b321123dfL62-L73) introduced to fix #44 caused a regression. 

By using Model.assembler() instead of Model.assembler(urlClassLoader) where applicable, we get rid of the calls to traitFactory and validatorFactory.

This result in a subtle change of behaviour where you don’t get validation from your external dependencies when working on your Smithy specification. It flew under the radar, and I caught completely by accident on an unrelated thing.

This original PR is intended to fail. It should serve as a point of discussion for a solution. I assume the test fails was added to cover a scenario that caused the initial #44. Rolling back would not be very wise.

I'd like to better understand why this happens for this particular example. When I do it'll be easier to come up with a fix that can actually work for everybody.

Specifically, I don't understand why, in #44, we have to case the result of the nodemapper deserializer. My understanding is that it is _this_ cast that makes the whole thing blow up.


daddykotex:
The content of the `TestTrait.class` contained in the jar looks like:

```
//
// Source code recreated from a .class file by IntelliJ IDEA
// (powered by FernFlower decompiler)
//

package software.amazon.smithy.test.traits;

import software.amazon.smithy.model.node.Node;
import software.amazon.smithy.model.node.NodeMapper;
import software.amazon.smithy.model.node.ObjectNode;
import software.amazon.smithy.model.shapes.ShapeId;
import software.amazon.smithy.model.traits.AbstractTrait;
import software.amazon.smithy.model.traits.AbstractTraitBuilder;
import software.amazon.smithy.model.traits.Trait;
import software.amazon.smithy.utils.ToSmithyBuilder;

public final class TestTrait extends AbstractTrait implements ToSmithyBuilder<TestTrait> {
    public static final ShapeId ID = ShapeId.from("smithy.test#test");

    private TestTrait(TestTrait.Builder builder) {
        super(ID, builder.getSourceLocation());
    }

    public static TestTrait.Builder builder() {
        return new TestTrait.Builder();
    }

    protected ObjectNode createNode() {
        NodeMapper mapper = new NodeMapper();
        mapper.disableToNodeForClass(TestTrait.class);
        mapper.setOmitEmptyValues(true);
        return mapper.serialize(this).expectObjectNode();
    }

    public TestTrait.Builder toBuilder() {
        return builder();
    }

    public static final class Builder extends AbstractTraitBuilder<TestTrait, TestTrait.Builder> {
        public Builder() {
        }

        public TestTrait build() {
            return new TestTrait(this);
        }
    }

    public static final class Provider extends software.amazon.smithy.model.traits.AbstractTrait.Provider {
        public Provider() {
            super(TestTrait.ID);
        }

        public Trait createTrait(ShapeId target, Node value) {
            TestTrait result = (TestTrait)(new NodeMapper()).deserialize(value, TestTrait.class);
            result.setNodeCache(value);
            return result;
        }
    }
}
```

daddykotex:
Oh I think that in the original source, there is no cast, but the JVM adds it when compiling to .class.

daddykotex:
One solution I've got is to tweak the cache in the `NodeMapper` so that the computed key takes into acount the class loader of the target type it's going to load:

```java
    private static ObjectCreatorFactory cachedCreator(ObjectCreatorFactory delegate) {
        ConcurrentMap<String, NodeMapper.ObjectCreator> cache = new ConcurrentHashMap<>();
        return (nodeType, target, nodeMapper) -> {
            int targetClassLoaderKey = ((Class) target).getClassLoader().hashCode(); // new
            String key = nodeType.getNodeClass() + ":" + target.getTypeName() + ":" + targetClassLoaderKey;
            return cache.computeIfAbsent(key, pair -> {
                return delegate.getCreator(nodeType, target, nodeMapper);
            });
        };
    }
```

This works but it means that in the context of a language server, it is essentially a memory leak: the server constantly reloads the model and its dependencies. This would mean that a key would only essentially be useful within a given load (one load a model using `TestTrait` twice, then the second time it's cached. But if one model loads it, then the language server reloads a model, we have a new class path, thus a new key).

milesziemer:
I think whats happening is that, in the language server, by creating a new URLClassLoader when the model is built each time, the trait is being loaded with two different class loaders so Java thinks the two TestTrait classes aren't equal when trying to do the cast. I could be wrong though.

I'm not sure making changes to smithy-model is the right solution here. Do you have a test case locally to reproduce not getting validation from external dependencies you can add to the PR? If not I can also make one. I have an idea I want to test.

daddykotex:
> two different class loaders so Java thinks the two TestTrait classes aren't equal when trying to do the cast

you're right, java fails because the same class comes from two different class loaders. But it happens because, on first time around it cache a `TestTrait` instance from class loader A, and then on the second time around, it looks at the target type, assumes it's the same thing and then try to do a cast. If you remove the cache, there is no problem.


----

I'll push an additional test case that uses a jar we publish here: https://github.com/disneystreaming/alloy

----

I've also tested just calling the `validatorFactory` of the assembler but it did not work, I guess it can't create the `trait` w/o the trait factory.

daddykotex:
In https://github.com/awslabs/smithy-language-server/pull/78/commits/6cf161e577b38bbe3591f8dff75d3158770518b0 I reverted and added a test to reproduce the error were getting on our side (or rather the lack of error should I say)

daddykotex:
Thanks for the feedback. I'll be on the lookout for your updates.

IMO, the problem happens because of a cache inside smithy-model, and that cache can't be disabled. I'll be happy if we can solve it locally by a having one url class loader that we reuse, but that sounds tricky.

milesziemer:
From my testing, we can pretty easily avoid creating an entirely new URLClassLoader each time we read the model. If we use a singleton URLClassLoader inside SmithyInterface, we can update it each time the model is read by using the previous URLClassLoader instance as the parent like this: 

`urlClassLoaderInstance = new URLClassLoader(urlArray, urlClassLoaderInstance)`

In the commit you mention, we stopped passing the URLClassLoader to the model assembler because it was causing errors when the same trait class was being loaded twice. The reason errors were happening was because the class loader used to load the trait class was different each time. By not including the class loader, we weren't getting the validators from the external jar even though they were present in the model. However, if we use the previous URLClassLoader as the parent class loader when we re load dependencies, the trait class isn't seen as a new class when we try to load it again, so there's no conflict. That means we should be able to revert back to including the class loader in the model assembler, and make sure we get validation from external jars.

My concern with this approach is having a huge "hierarchy" of class loaders, so I'll have to look into that more. I was able to test adding and removing external dependencies successfully, but I also want to do some general testing and performance testing too.

milesziemer:
Just wanted to give an update on where I'm at with this. I wasn't able get my idea of using a single URLClassLoader to work, I'm pretty much out of ideas for how to keep the changes in the language server. Your idea of tweaking the caching in smithy-model seems to be the way to go right now, and we're working on a way to modify the behavior of the cache to always keep the most recently loaded version of the class, evicting the old version when necessary. Still needs some more testing, but I'll keep you updated.

daddykotex:
Unfortunate :(

A way to disable the cache would work too.

milesziemer:
An update on the status of this issue: we've just merged https://github.com/awslabs/smithy/pull/1518 in Smithy, which modifies caching to solve this issue. It'll be available in the next Smithy release. Once that's out, we can update the language server's Smithy dependency.

daddykotex:
> An update on the status of this issue: we've just merged [awslabs/smithy#1518](https://github.com/awslabs/smithy/pull/1518) in Smithy, which modifies caching to solve this issue. It'll be available in the next Smithy release. Once that's out, we can update the language server's Smithy dependency.

I've reverted the last commit in this PR. The changes you made in Smithy (which I published locally) works fine. I'll do a e2e release (smithy + the language server) and give it a run in VSCode to see if my original issue is resolved. I'll report back then.

Thanks for your hard work

daddykotex:
It works. Thanks again, we can re-trigger this PR workflow when smithy 1.26.4 is released and that should do it.

milesziemer:
We released the fixes in Smithy in 1.27.0, and released the language server with an updated dependency. I'm going to close this now since it is fixed.

PR #80, opened 2022-12-08
Add traits, validators from upstream

kubukoz:
*Description of changes:*

Overrides the trait and validator factories to include instances captured by SPI from `externalJars`.

---

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
@@ -24,7 +24,9 @@ import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import software.amazon.smithy.lsp.ext.LspLog;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.loader.ModelAssembler;
+import software.amazon.smithy.model.traits.TraitFactory;
 import software.amazon.smithy.model.validation.ValidatedResult;
+import software.amazon.smithy.model.validation.ValidatorFactory;
 
 public final class SmithyInterface {
   private SmithyInterface() {
--- a/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
@@ -46,28 +46,29 @@ public final class SmithyInterface {
       Model.Builder builder = Model.builder();
 
       URL[] urlArray = externalJars.stream().map(SmithyInterface::fileToUrl).toArray(URL[]::new);
+      URLClassLoader urlClassLoader = new URLClassLoader(urlArray);
 
-      if (urlArray.length > 0) {
-        // Loading the model just from upstream dependencies, in isolation, and adding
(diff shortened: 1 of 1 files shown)


src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
```diff
@@ -46,29 +46,30 @@ public static Either<Exception, ValidatedResult<Model>> readModel(Collection<Fil
       Model.Builder builder = Model.builder();
 
       URL[] urlArray = externalJars.stream().map(SmithyInterface::fileToUrl).toArray(URL[]::new);
+      URLClassLoader urlClassLoader = new URLClassLoader(urlArray);
 
-      if (urlArray.length > 0) {
```
kubukoz, line 50: I can revert this particular chunk, as the only thing I really need outside of it is the `urlClassLoader` value, but I think the condition is redundant: if the array is empty, we'll just add "no shapes".

src/main/java/software/amazon/smithy/lsp/SmithyInterface.java
```diff
@@ -46,29 +46,30 @@ public static Either<Exception, ValidatedResult<Model>> readModel(Collection<Fil
...
+      builder.addShapes(upstreamModel);
 
       ModelAssembler assembler = Model.assembler()
               // We don't want the model to be broken when there are unknown traits,
               // because that will essentially disable language server features.
               .putProperty(ModelAssembler.ALLOW_UNKNOWN_TRAITS, true)
               .addModel(builder.build());
 
+      assembler.validatorFactory(ValidatorFactory.createServiceFactory(urlClassLoader));
+      assembler.traitFactory(TraitFactory.createServiceFactory(urlClassLoader));
```
kubukoz, line 71: Felt right, but if you think it's problematic - let me know :)

kubukoz: ah, lovely:

> `java.lang.ClassCastException: class software.amazon.smithy.test.traits.TestTrait cannot be cast to class software.amazon.smithy.test.traits.TestTrait (software.amazon.smithy.test.traits.TestTrait is in unnamed module of loader java.net.URLClassLoader @2d7275fc; software.amazon.smithy.test.traits.TestTrait is in unnamed module of loader java.net.URLClassLoader @b2c5e07)`

milesziemer:
Thanks for contributing, I think this PR inadvertently reverts a change made in https://github.com/awslabs/smithy-language-server/pull/44, that fixed the error you are seeing. Can you elaborate on what you're trying to do with this change?

kubukoz:
I'm using a library that publishes Smithy files, along with some trait classes and validators for them. In the project that uses these (by referencing an artifact in `smithy-build.json`), I want the validators to run, and for any diagnostics produced by that validation - display them in the IDE.

It seems to work for my usecase, but clearly it's hitting other issues :)

kubukoz:
Sounds like it! I'll check that one out :)

kubukoz:
will do, thanks. I think we can close this, then.

PR #86, opened 2023-01-12
Diagnostic and codeActions for missing/outdated Smithy version

daddykotex:
We were wondering what would be a good way to push users towards adding the `$version` statement in their Smithy file. By default, if it's missing, smithy-model assumes `1`. But most Smithy files could be using `2` so we discussed a way to have this in the IDE.

The flow is basically: if you don't have a version, we suggest/help to define it explicitly (to 1), and if you use version 1, we suggest/help to update it to 2.

The end result looks like this in VS Code:

https://user-images.githubusercontent.com/5230460/212099585-8f00378d-e1b5-495a-b3f2-c78b2cca5ebf.mov

It works by:

1. publishing a diagnostic if the version is missing
2. publishing a diagnostic if the version is found but it contains an older version (`1` here)
3. implement codeAction to generate `TextEdit` if the diagnostic above are found

For 1, and 2, we read the top part of the Smithy file (in this PR I read 5 lines but this is an arbitrary amount). Then we look for a line starting with `$version`. If not found we publish a diagnostic, if found but equal to 1, we publish a diagnostic.

Those diagnostic are added on top of any diagnostic generated from validating the model.

The `TextEdit` are simple:
- rewrite the version if it is found (we can reuse the range from the diagnostic to build the `TextEdit`)
- append `$version: "1"` at the top of the file, explicitly setting the version for the user

------

On top of that, this PR includes (those could be separated PRs):

- a typo fix in some java doc
- a fix when using `findToken` on an empty document

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
@@ -34,8 +34,12 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.eclipse.lsp4j.Command;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.CodeActionParams;
 import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.CompletionList;
 import org.eclipse.lsp4j.CompletionParams;
@@ -61,6 +65,9 @@ import org.eclipse.lsp4j.TextDocumentItem;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.services.LanguageClient;
 import org.eclipse.lsp4j.services.TextDocumentService;
+import software.amazon.smithy.lsp.codeactions.DefineVersionCodeAction;
+import software.amazon.smithy.lsp.codeactions.SmithyCodeActions;
+import software.amazon.smithy.lsp.codeactions.UpdateVersionCodeAction;
 import software.amazon.smithy.lsp.ext.Completions;
 import software.amazon.smithy.lsp.ext.Constants;
 import software.amazon.smithy.lsp.ext.Document;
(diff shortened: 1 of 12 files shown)


srchase:
This is great.

What's the rational for defaulting to `1` instead of `2` when no version is defined?

It looks like the `1` to `2` upgrade is suggested regardless of whether the model will work as a `2` model. Is the intent that they will get diagnostics on anything that isn't compatible with `2` and be able to make the other needed changes on their own?

It'd be awesome if the CodeAction could handle making more of those changes with a preview of what it would change, but maybe that could be added later. I think at a minimum, it'd be helpful to have the `1` to `2` suggestion include some additional context. Perhaps link to the [migration guide](https://smithy.io/2.0/guides/migrating-idl-1-to-2.html)?

daddykotex:
> It'd be awesome if the CodeAction could handle making more of those changes with a preview of what it would change, but maybe that could be added later. I think at a minimum, it'd be helpful to have the 1 to 2 suggestion include some additional context. Perhaps link to the [migration guide](https://smithy.io/2.0/guides/migrating-idl-1-to-2.html)?

Do you mean that it could be valuable to have a custom action to update all files (to insert a version, or update a version)?

> I think at a minimum, it'd be helpful to have the 1 to 2 suggestion include some additional context.

Good idea, I'll see if I can add that

srchase:
> Do you mean that it could be valuable to have a custom action to update all files (to insert a version, or update a version)?

I meant that for a given file, the code action which is changing `1` to `2` would also be able to do the `1` to `2` migration on that file. For example, it could detect set shapes and change them to list shapes.

daddykotex:
> > Do you mean that it could be valuable to have a custom action to update all files (to insert a version, or update a version)?
> 
> I meant that for a given file, the code action which is changing `1` to `2` would also be able to do the `1` to `2` migration on that file. For example, it could detect set shapes and change them to list shapes.

Ahhh ok I see what you mean. I think I prefer to scope code action per diagnostic. In practice that means that once you move to version 2. You'll get a new set of diagnostic. If you use `set` you'd get one for that.  We could have an independent code action that does that: turn `set` into `uniqueItems` on `list`. Similarly, we could have a codeaction to rewrite `enum` I believe.


> I think at a minimum, it'd be helpful to have the 1 to 2 suggestion include some additional context.

Regarding the note, I think I can add some context to the diagnostic, but not the code action. See the following examples from the Shellcheck extension:

Screen Shot 2023-01-12 at 13 47 34

Screen Shot 2023-01-12 at 13 47 27

srchase:
> Regarding the note, I think I can add some context to the diagnostic, but not the code action. See the following examples from the Shellcheck extension:

Adding to the diagnostic sounds good.

daddykotex:
Implemented the `href` code description:

Screen Shot 2023-01-12 at 15 20 12

It requires an update on the client side because it's only supported as of LSP 3.16.0

src/main/java/software/amazon/smithy/lsp/codeactions/DefineVersionCodeAction.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
srchase, line 2: nit: wrong year on all of these headers.

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -510,6 +535,34 @@ private List<ValidationEvent> getValidationEventsForShape(Shape shape) {
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public CompletableFuture<List<Either<Command, CodeAction>>> codeAction(CodeActionParams params) {
```
milesziemer, line 539: I wonder if we could take the implementation of this method and stick it somewhere in the new codeactions folder, maybe in SmithyCodeActions.java? That way, we can more easily test the code actions in isolation, and not add more bloat to the text document service class.

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -332,8 +345,11 @@ private List<String> textBufferContents(String path) throws IOException {
         return contents;
     }
 
-    private String findToken(String path, Position p) throws IOException {
+    private Optional<String> findToken(String path, Position p) throws IOException {
```
srchase, line 348: Can this fix for the `findToken` on empty documents be moved to its own PR? This PR is trying to do a lot with this bundled in with the CodeAction changes.

srchase (changes requested):

milesziemer:
This is awesome, thanks for doing it. Would you mind splitting up the other changes you mentioned into separate PR's so it is easier to review?

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
daddykotex: yeah, do you mind if rebase to remove the relevant commits?

srchase: That works.

daddykotex:
Refactored code action and diagnostic for the version in their own file and created a test suite for the whole version refactoring feature (code action and diagnostic for version)

srchase (approved):

PR #95, opened 2023-03-25
Add github release task

RenFraser:
*Description of changes:*

Adds a GitHub action to create a release each time a new tag is pushed. This change is made with the intention of making the language server more accessible by removing the need to build the artifacts if users wish. 

For my use case, this will allow me to add the language server to [Homebrew](https://brew.sh).

Because the workflow is triggered on tags prepended with `v`, for instance `v1.2.3`. I can see that there aren't any tags in the repo, so let me know if that's not an option and perhaps we could discuss a path forward :)

*Testing*

See my [run of the deploy job](https://github.com/RenFraser/smithy-language-server/actions/runs/4517612098/jobs/7956939050) where I pushed tags `vtest` and [associated release](https://github.com/RenFraser/smithy-language-server/releases/tag/vtest)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,36 @@
+---
+
+name: deploy
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  deploy:
+    permissions:
+      contents: write # needed for the release script
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set variables
+        run: |
+          VER=$(cat VERSION)
+          echo "VERSION=$VER" >> $GITHUB_ENV
+          echo "Version is $VER"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
+---
+
+name: deploy
+on:
+  push:
+    tags:
+      - "v*"
```
srchase, line 7: Can we prefix with `0` instead? I think the only tags will be for releases, but we can at least filter on the 0.x.y tags. When we hit 1.x.y, we can revisit if releasing on every tag makes sense or not.

RenFraser: Thanks for the feedback @srchase, I've pushed a commit to change it from `v*` to `0*` as you've requested :)

srchase:
Thanks for the contribution!!

Issue #97, opened 2023-04-14
Workflows Failing


nickrodgers42:
I was trying to build the language server using the `.gradlew` file and noticed that it was failing on my machine (WSL 2 using Ubuntu). Given that it's a weird platform I tried on my Macbook, same failures. I forked the repo and ran the workflows on the tip of `main`, only changing the .README file, same failures. 

https://github.com/nickrodgers42/smithy-language-server/actions/runs/4704512621/jobs/8344210410

srchase:
This looks like it started with the release of Smithy `1.30.0`.

If I pin the Smithy version to `1.29.0` in build.gradle ([here](https://github.com/awslabs/smithy-language-server/blob/main/build.gradle#L160), the running `./gradlew clean build` succeeds.

I need look into what specific change in `1.30.0` is causing this issue on the Language Server.

Due to the `definitionLocationsV2` test failing, it's likely the cause of https://github.com/awslabs/smithy-language-server/issues/96.

srchase:
Resolved in #98.

PR #98, opened 2023-04-18
Fix location for elided members

srchase:
The change in https://github.com/awslabs/smithy/pull/1715 which was released with Smithy [1.30.0](https://github.com/awslabs/smithy/releases/tag/1.30.0) fixed the source location of members which are mixed-in, but not re-defined within a structure. This fix resulted in builds for the Language Server failing, as noted in https://github.com/awslabs/smithy-language-server/issues/97.

This PR updates the tests and logic to determine the proper location for mixed-in members that are not re-defined within a structure. Instead of a zero-length location at the start of the structure, members which are not re-defined within the containing structure will instead use the location of the mixed-in member.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
+++ b/src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
@@ -51,6 +51,7 @@ final class FileCachingCollector implements ShapeLocationCollector {
     private Map<String, ModelFile> fileCache;
     private Map<OperationShape, List<Shape>> operationsWithInlineInputOutputMap;
     private Map<ShapeId, List<MemberShape>> containerMembersMap;
+    private Map<ShapeId, ShapeId> membersToUpdateMap;
 
     @Override
     public Map<ShapeId, Location> collectDefinitionLocations(Model model) {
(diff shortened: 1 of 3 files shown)


mtdowling (approved):

src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
```diff
@@ -59,13 +60,16 @@ public Map<ShapeId, Location> collectDefinitionLocations(Model model) {
...
         this.containerMembersMap = new HashMap<>();
+        this.membersToUpdateMap = new HashMap<>();
 
         for (ModelFile modelFile : this.fileCache.values()) {
             collectContainerShapeLocationsInModelFile(modelFile);
         }
 
         operationsWithInlineInputOutputMap.forEach((this::collectInlineInputOutputLocations));
         containerMembersMap.forEach(this::collectMemberLocations);
+        // Make final pass to set locations for mixed-in member locations that weren't available on first past.
```
syall, line 71: nit: typo

```suggestion
        // Make final pass to set locations for mixed-in member locations that weren't available on first pass.
```

syall (approved):

PR #101, opened 2023-04-24
Fix off by one issue

### LINKED ISSUE #100: IndexOutOfBoundsException when loading project
If the project contains a file where:

- an operation input or output is defined "outline" (as in, not inline)
- the shape in question occupies only a single line in the file
- it's the last line of the file

The project load fails due to an out-of-bounds exception (e.g. index 17 in a 17-element list).

Minimal file to reproduce:

```smithy
$version: "2"

namespace demo

operation MyOp {
    output: MyOpOutput
}

@output
structure MyOpOutput {}
```

I have a fix in the works.

kubukoz:
*Description of changes:*

Fixes an issue preventing the load of a valid Smithy workspace. Closes #100.


src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
```diff
@@ -274,7 +274,9 @@ private Boolean shapeWasDefinedInline(OperationShape operation, Shape shape, Mod
         }
         int boundary = Math.max(priorShapeLine, operation.getSourceLocation().getLine());
         while (shapeStartLine >= boundary) {
-            String line = modelFile.lines.get(shapeStartLine);
+            String line = modelFile.lines.get(shapeStartLine - 1);
+
+            // note: this doesn't take code inside comments into account
```
kubukoz, line 279: This might be an issue but I haven't gotten around to provoking it yet.

src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
```diff
@@ -274,7 +274,9 @@ private Boolean shapeWasDefinedInline(OperationShape operation, Shape shape, Mod
         }
         int boundary = Math.max(priorShapeLine, operation.getSourceLocation().getLine());
         while (shapeStartLine >= boundary) {
-            String line = modelFile.lines.get(shapeStartLine);
+            String line = modelFile.lines.get(shapeStartLine - 1);
```
kubukoz, line 277: I'm not 100% sure this is the right thing to do, but I checked around to confirm, and it seems like there's usually a `-1` happening when we look up lines in other places. I do think this is the right behavior.

src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
srchase: Are you planning to test further to see if this is an issue?

srchase:
> @srchase can you please take another look? I think the job failed due to an intermittent failure.

Thanks for fixing this!

PR #104, opened 2023-05-01
Upgrade to Smithy 1.31.0 and update Hover provider

srchase:
Upgrades the Language Server to use 1.31.0.

The Hover provider uses [SmithyIdlModelSerializer](https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java) to generate a serialized preview of the target shape. Previously, if the target shape had validation, the serializer would throw an exception. Beginning with 1.31.0, the serializer is able to successfully serialize these shapes.

This PR updates the Hover provider to append validation events to the returned hover content.

In the attached screenshot, the target `Foo` shape is being hovered over within the `Bar` structure. `Foo`, defined in foo.smithy, has an unknown trait applied and the hover now shows the serialized shape, along with the validation `WARNING` about the unknown trait that has been applied to that shape.

hover

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -157,7 +157,7 @@ publishing {
 
 dependencies {
     implementation "org.eclipse.lsp4j:org.eclipse.lsp4j:0.14.0"
-    implementation "software.amazon.smithy:smithy-model:[1.30.0, 2.0["
+    implementation "software.amazon.smithy:smithy-model:[1.31.0, 2.0["
     implementation 'io.get-coursier:interface:1.0.4'
     implementation 'com.disneystreaming.smithy:smithytranslate-formatter-jvm-java-api:0.3.4'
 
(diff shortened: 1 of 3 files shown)


mtdowling (approved):

build.gradle
```diff
@@ -157,7 +157,7 @@ publishing {
 
 dependencies {
     implementation "org.eclipse.lsp4j:org.eclipse.lsp4j:0.14.0"
-    implementation "software.amazon.smithy:smithy-model:[1.30.0, 2.0["
+    implementation "software.amazon.smithy:smithy-model:[1.31.0, 2.0["
```
kubukoz, line 160: hey @srchase I'm wondering what this syntax means: `[1.31.0, 2.0[` - I understand `[1.31.0` means we need _at least 1.31.0 including that version_, but the `2.0[` part puzzles me. Any hints?

srchase: You got it. That syntax is a little funky.

PR #105, opened 2023-05-10
More verbosity in logs/output panel in case of errors

kubukoz:
*Description of changes:*

While fighting some issues in a project not loading (recently seeing some similar to #100), I found that we were discarding some useful information, such as stack traces, which makes it more difficult to debug issues on remote machines (such as my colleagues').

This PR adds more `e.printStackTrace()` in cases where we only printed the exception _messages_. While this isn't perfect, and it's far from ideal UX when you see potentially duplicate stack traces, I find that in this case more is better than none.

Many of the issues I'm seeing anytime something goes wrong in loading a project are NPEs on the `project` field in `SmithyTextDocumentService`, so I also marked that as nullable to make it clearer that it's not a safe field to use in all cases. We can amend the usages of that field in future PRs so that they have reasonable fallbacks.

Example of what would happen after this change given some issue in `FileCachingCollector` (responsible for things like #100):

image

---

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
@@ -82,7 +82,7 @@ public class SmithyLanguageServer implements LanguageServer, LanguageClientAware
         tempWorkspaceRoot.deleteOnExit();
         WorkspaceFolder workspaceFolder = new WorkspaceFolder(tempWorkspaceRoot.toURI().toString());
         params.setWorkspaceFolders(ListUtils.of(workspaceFolder));
-      } catch (Exception e) {
+      } catch (IOException e) {
         e.printStackTrace();
       }
     }
--- a/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
@@ -69,7 +69,7 @@ public class SmithyLanguageServer implements LanguageServer, LanguageClientAware
         loadSmithyBuild(workspaceRoot);
       } catch (Exception e) {
         LspLog.println("Failure trying to load extensions from workspace root: " + workspaceRoot.getAbsolutePath());
-        LspLog.println(e.toString());
+        e.printStackTrace();
       }
     } else {
       LspLog.println("Workspace root was null");
@@ -78,11 +78,11 @@ public class SmithyLanguageServer implements LanguageServer, LanguageClientAware
     if (params.getWorkspaceFolders() == null) {
       try {
         tempWorkspaceRoot = Files.createTempDirectory("smithy-lsp-workspace").toFile();
-        System.out.println("Created temporary workspace root: " + tempWorkspaceRoot);
+        LspLog.println("Created temporary workspace root: " + tempWorkspaceRoot);
         tempWorkspaceRoot.deleteOnExit();
         WorkspaceFolder workspaceFolder = new WorkspaceFolder(tempWorkspaceRoot.toURI().toString());
         params.setWorkspaceFolders(ListUtils.of(workspaceFolder));
-      } catch (IOException e) {
+      } catch (Exception e) {
         e.printStackTrace();
       }
     }
@@ -95,6 +95,7 @@ public class SmithyLanguageServer implements LanguageServer, LanguageClientAware
         loadSmithyBuild(root);
       } catch (Exception e) {
         LspLog.println("Error when loading workspace folder " + ws.toString() + ": " + e);
+        e.printStackTrace();
       }
     }
 
--- a/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
@@ -38,6 +38,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.Nullable;
 import org.eclipse.lsp4j.CodeAction;
 import org.eclipse.lsp4j.CodeActionParams;
 import org.eclipse.lsp4j.Command;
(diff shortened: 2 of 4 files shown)


src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -142,12 +144,17 @@ public Optional<File> getRoot() {
     public void createProject(SmithyBuildExtensions ext, File root) {
         Either<Exception, SmithyProject> loaded = SmithyProject.load(ext, root);
         if (loaded.isRight()) {
-            this.project = loaded.getRight();
+            SmithyProject project = loaded.getRight();
```
kubukoz, line 147: **thought:** this local is only here to avoid redundant warnings in IDEs saying that `this.project` is nullable in the lines below.

src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java
```diff
@@ -438,7 +450,7 @@ public CompletableFuture<List<Either<SymbolInformation, DocumentSymbol>>> docume
                     .collect(Collectors.toList())
             );
         } catch (Exception e) {
-            e.printStackTrace(System.err);
```
kubukoz, line 441: **issue:** This is the default. It's on me, I added this line ;)

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -172,7 +172,13 @@ private static Either<Exception, SmithyProject> load(
             return Either.forLeft(model.getLeft());
         } else {
             model.getRight().getValidationEvents().forEach(LspLog::println);
-            return Either.forRight(new SmithyProject(imports, smithyFiles, externalJars, root, model.getRight()));
+
+            try {
+                SmithyProject p = new SmithyProject(imports, smithyFiles, externalJars, root, model.getRight());
```
kubukoz, line 177: **thought:** The issues I've been seeing are all during this constructor call, so there's no reason to assume it doesn't throw. Catching here means that `createProject` will send a message to the client. Potentially it'd be a better place for the try/catch? Or should we do it in both, just for future generations?

src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
```diff
@@ -63,7 +63,12 @@ public Map<ShapeId, Location> collectDefinitionLocations(Model model) {
         this.membersToUpdateMap = new HashMap<>();
 
         for (ModelFile modelFile : this.fileCache.values()) {
-            collectContainerShapeLocationsInModelFile(modelFile);
+            try {
+                collectContainerShapeLocationsInModelFile(modelFile);
+            } catch (Exception e) {
+                throw new RuntimeException("Exception while collecting container shape locations in model file: "
```
kubukoz, line 69: **thought:** Just wrapping the underlying one to add the extra information about the problematic file. This might make it easier to pinpoint the piece of syntax that the file collector is struggling with.

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -69,7 +69,7 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
         loadSmithyBuild(workspaceRoot);
       } catch (Exception e) {
         LspLog.println("Failure trying to load extensions from workspace root: " + workspaceRoot.getAbsolutePath());
-        LspLog.println(e.toString());
+        e.printStackTrace();
```
milesziemer, line 72: Does this create the same issue you described in [this comment](https://github.com/awslabs/smithy-language-server/pull/105/files#r1189255607)? Does printing to System.err mess with LSP transport?

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -78,11 +78,11 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
     if (params.getWorkspaceFolders() == null) {
       try {
         tempWorkspaceRoot = Files.createTempDirectory("smithy-lsp-workspace").toFile();
-        System.out.println("Created temporary workspace root: " + tempWorkspaceRoot);
+        LspLog.println("Created temporary workspace root: " + tempWorkspaceRoot);
         tempWorkspaceRoot.deleteOnExit();
         WorkspaceFolder workspaceFolder = new WorkspaceFolder(tempWorkspaceRoot.toURI().toString());
         params.setWorkspaceFolders(ListUtils.of(workspaceFolder));
-      } catch (IOException e) {
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
milesziemer: This catches IOException because the `Files.createTempDirectory` method throws a checked IOException. I don't think we should catch more here, a different error means there's something else wrong and we shouldn't try to recover from that.

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
kubukoz: it does not, system.err is fine and gets piped to the output panel (at least in vscode-languageclient) :)
only stdin/stdout are a problem

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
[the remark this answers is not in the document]
kubukoz: fair enough, happy to revert this change

milesziemer:
Looking at this PR makes me think we need to change the way SmithyProject is being used inside SmithyTextDocumentService. I feel like the NPEs are coming from the fact that [createProject](https://github.com/awslabs/smithy-language-server/blob/8b31c022effe03a6dc01ec22104468f60f2310d5/src/main/java/software/amazon/smithy/lsp/SmithyTextDocumentService.java#L142) won't set the `project` property if the project fails to load, so anywhere else we use `project` will just get an NPE, and then we have try-catch everywhere. We should think about what functionality is offered if the SmithyProject fails to load, and maybe split that up somehow so we have more deterministic behavior.

kubukoz:
I agree in principle, but that's a bigger change and I wanted to provide users with a little bit more context for troubleshooting in the meantime.

milesziemer (approved):
Fair enough, I agree it makes sense to provide people with the extra debug info for now while the language server is still under development. We can circle back and work on how the LSP communicates with the user when we iron out some more of the rough edges.

PR #108, opened 2023-05-23
Add WebSocket support and launcher

haydenbaker:
*Description of changes:*
- Add websocket support for LSP, such that using the LSP over websockets is possible

*Testing:*
- Launch WSS via `./gradlew run --args="3000 --ws"`, confirm a websocket client can connect and communicate with it
- Launch stdin/stdout via `./gradlew run --args="0"`, confirm no behavior change
- Bump version, Publish to maven local, update vscode extension version, confirm no behavior change


src/main/java/software/amazon/smithy/lsp/Main.java
```diff
@@ -61,21 +65,28 @@ public static void main(String[] args) {
         Socket socket = null;
         InputStream in;
         OutputStream out;
-
+        List<String> argList = Arrays.asList(args);
         try {
-            String port = args[0];
+            Optional<Exception> launchFailure;
+            String port = getOrDefault(argList, 0, "0");
+            String type = getOrDefault(argList, 1, null);
```
justinmk3, line 72: to avoid order sensitivity it may be worth taking a dependency on https://github.com/argparse4j/argparse4j or similar

haydenbaker: Yeah, I didn't really want to make this a full blown CLI with argparse, but maybe something we should do in the future if we decide to add more. Great suggestion.

I also just noticed this package doesn't have a great readme, so I'll add some details there (and include the ordering).

src/main/java/software/amazon/smithy/lsp/Main.java
```diff
@@ -61,21 +65,28 @@ public static void main(String[] args) {
...
             // If port is set to "0", use System.in/System.out.
             if (port.equals("0")) {
                 in = System.in;
                 out = System.out;
+                launchFailure = launch(in, out);
+            } else if (type != null && type.equals("--ws")) {
```
milesziemer, line 78:
```suggestion
            } else if ("--ws".equals(type)) {
```
Minor, don't have to do the null check like this.

src/main/java/software/amazon/smithy/lsp/Main.java
milesziemer, line 72: If the websocket path returns right away, maybe we can hoist that out of the if-else into a guard clause, and not change the existing if-else.
```suggestion
            String port = getOrDefault(argList, 0, "0");
            String type = getOrDefault(argList, 1, null);
            
            if ("--ws".equals(type)) {
              // Do websocket stuff
              return;
            }
            
            // Rest is unchanged
```

milesziemer (changes requested):

milesziemer (changes requested):

README.md
```diff
@@ -4,6 +4,36 @@
...
+Run `./gradlew run --args="12423"`
+
+The LSP will try to connect to the given port using a TCP socket - if it can't, it will fail.
+
+This is used by the VSCode extension to establish a connection between it and the LSP (which is launched
+as a local process)
+
+#### WebSockets
+
+Run `./gradlew run --args="3000 --ws"`
```
mtdowling, line 31: Can you put `--ws` in front? I would assume here that `3000` is a positional argument that comes after named arguments, and those are usually last.

kubukoz: perhaps it's also worth considering the deprecation of `port` as a positional argument and making it a `--port` option? For some time the server could support both styles, but eventually the positional way would be dropped.

PR #111, opened 2023-05-29
Fix matching inline inputs/outputs with operations

### LINKED ISSUE #110: IndexOutOfBoundsException preventing project load when a naming conflict occurs
To reproduce:

1. Create a file, `a.smithy`, with this content:

```smithy
$version: "2"

namespace a

operation HelloWorld {
    input := {
        @required
        name: String
    }
    output := {
        @required
        name: String
    }
}

```

2. Create another one, `b.smithy`, with this content:

```smithy
$version: "2"

namespace b

string Ignored

operation HelloWorld {
    input := {
        @required
        name: String
    }
    output := {
        @required
        name: String
    }
}
```

3. See errors. In your extension output panel you may see this:
```
java.lang.IndexOutOfBoundsException: Index 15 out of bounds for length 14
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:266)
	at java.base/java.util.Objects.checkIndex(Objects.java:359)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at software.amazon.smithy.lsp.ext.FileCachingCollector.getNextEndMarker(FileCachingCollector.java:362)
	at software.amazon.smithy.lsp.ext.FileCachingCollector.getEndPosition(FileCachingCollector.java:421)
	at software.amazon.smithy.lsp.ext.FileCachingCollector.collectInlineInputOutputLocations(FileCachingCollector.java:153)
	at software.amazon.smithy.lsp.ext.FileCachingCollector.lambda$collectDefinitionLocations$0(FileCachingCollector.java:79)
```

I debugged this a little bit while minimizing, and it's pretty obvious to me now that the problem is the `HelloWorld` naming conflict - somewhere in the file caching collector there's some logic that doesn't disambiguate the two operations, while it should.

I'll investigate this further and work on a fix.

kubukoz:
*Issue #, if available:* #110

*Description of changes:*

This fixes #110 (an IIOBE) by preventing the underlying issue: the `FileCachingCollector` mixing up operations with the same name in different namespaces.

Unfortunately, the test that reproduced the issue is flaky, as it would only fail about half the time before the fix was added - I think there's some non-deterministic behavior in some part of the collector that affects the order of things being collected, and sometimes prevents exhibiting the incorrect behavior. I've confirmed that it reliably passes now after the fix was applied.

### DIFF
--- a/src/test/resources/software/amazon/smithy/lsp/ext/models/operation-name-conflict/b.smithy
+++ b/src/test/resources/software/amazon/smithy/lsp/ext/models/operation-name-conflict/b.smithy
@@ -1,0 +1,16 @@
+$version: "2"
+
+namespace b
+
+string Ignored
+
+operation HelloWorld {
+    input := {
+        @required
+        name: String
+    }
+    output := {
+        @required
+        name: String
+    }
+}
--- a/src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
+++ b/src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
@@ -239,9 +239,12 @@ final class FileCachingCollector implements ShapeLocationCollector {
         ) {
             String suffix = getOperationInputOrOutputSuffix(shape, preamble);
             String shapeName = shape.getId().getName();
+
             String matchingOperationName = shapeName.substring(0, shapeName.length() - suffix.length());
+            ShapeId matchingOperationId = ShapeId.fromParts(shape.getId().getNamespace(), matchingOperationName);
+
             return model.shapes(OperationShape.class)
-                    .filter(operationShape -> operationShape.getId().getName().equals(matchingOperationName))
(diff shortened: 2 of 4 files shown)


src/test/resources/software/amazon/smithy/lsp/ext/models/operation-name-conflict/b.smithy
```diff
@@ -0,0 +1,16 @@
+$version: "2"
+
+namespace b
+
+string Ignored
```
kubukoz, line 5: **thought:** this technically can be anything that makes this file longer than the other - which is what ultimately caused the IIOBE (line lookup hitting a different file than intended, which would go over the bounds).

PR #118, opened 2023-08-28
Refactor hover and definition tests

milesziemer:
- Moves tests for hover and definition into their own test files.
- Refactors Harness to use a builder instead of overloaded `create` methods, and updates other tests to use the builder.
- Moves StubClient into its own file to be used by hover and definition tests.
- Adds TestUtils class to house common methods used by tests.
- Adds `setProject` method to SmithyTextDocumentService for usage by tests.

Eventually I think we can simplify the v1 vs v2 tests more since some of them are shared, and make these kinds of tests parameterized with junit5.

*Issue #, if available:*

*Description of changes:*


src/test/java/software/amazon/smithy/lsp/TestUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import junit.framework.AssertionFailedError;
+import org.eclipse.lsp4j.Location;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import software.amazon.smithy.lsp.ext.SmithyProjectTest;
+
+import static org.junit.Assert.assertTrue;
```
mtdowling, line 12: Not necessary, but FYI you can actually make custom JUnit assertions. See for example https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/test/java/software/amazon/smithy/model/ShapeMatcher.java

mtdowling (approved):

src/test/java/software/amazon/smithy/lsp/TestUtils.java
milesziemer: I'd like to move the language server to hamcrest and junit5 like the Smithy repo, maybe this can be a part of that. Seems useful

PR #120, opened 2023-08-29
Add workflow to automate PR for bumping Smithy

srchase:
This PR adds a workflow which checks for new Smithy versions and creates a PR to update the Language Server with that new version.


AndrewFossAWS (approved):

kubukoz:
this is good, but have you considered using something more generic to automate bumps of gradle dependencies in general?

PR #121, opened 2023-09-01
Store model text in memory

milesziemer:
This is primarily a refactor of SmithyProject to store raw model text in memory. Previously, we only stored file objects for the model, and gave those to the model assembler. There are two issues with this approach:
1. While a file is being editted, the source of truth for the file contents should be what the client sends back to the server in the `didChange` event, not whats actually in the file. To get around this, we've been using a temporary file that stores the contents of the file being editted, and giving that to the model assembler. When publishing diagnostics however, we needed to re-map the file location from the temporary file to the actual file on disk.
2. The language server needs literal text to provide some of its features, meaning we need to read in files frequently.

These changes update SmithyProject to store a map of URI -> String for each of the model files in the loaded model by walking the shapes and collecting all unique file locations. SmithyTDS is updated to use that when it needs to have literal text of the model for formatting, version diagnostics, and other features.

Various other updates were also made:
- Added a list of errors to SmithyProject that store any errors that occur while loading the project. Previously, any "load" methods would return `Either<Exception, SmithyProject>`, which isn't very ergonomic to use. These errors represent failures in the project loading, not validation errors from loading the model.
- SmithyProject's loading methods are reworked to provide 2 static methods: one for loading the project in a directory, one for loading from specific smithy-build config, and 2 instance methods: one for regular reloading, one for reloading with specific file changes. Previously, we had to manage the loading of the temporary file (or its removal). This also moves all the loading logic into SmithyProject.
- String-literal uris are replaced with URI class since LSP guarantees uris will be valid (see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri).
- SmithyInterface has new method to load from text-literal sources.
- Completions updated to handle invalid model result, so callers don't have to. This allows removing `getCompletions` from SmithyProject.
- Various access modifiers adjusted to be more strict, tightening the API to make it easier for us to change implementations.

The following test updates were made:
- Added multiple test cases for different project loading & reloading situations.
- Some tests needed to use URI class instead of uri strings.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
+++ b/src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
@@ -17,14 +17,14 @@ package software.amazon.smithy.lsp.ext;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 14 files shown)


src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -38,87 +38,158 @@
...
+     * @param contents Contents of the changed file.
+     * @return The reloaded project.
+     */
+    public SmithyProject reloadWithChanges(URI changedUri, String contents) {
+        this.modelFiles.put(changedUri, contents);
+        // Reload the model using in-memory versions of source files
+        List<File> existingSourceFiles = onlyExistingFiles(this.smithyFiles);
+        // Handle the case when the file is new
+        if (!existingSourceFiles.contains(new File(changedUri))) {
+            existingSourceFiles.add(new File(changedUri));
```
JordonPhillips, line 177: nit: introduce variable rather than constructing the file twice

src/main/java/software/amazon/smithy/lsp/ext/SmithyProject.java
```diff
@@ -146,63 +217,134 @@ public static Either<Exception, SmithyProject> load(SmithyBuildExtensions config
...
-        return SmithyInterface.readModel(discoveredFiles, externalJars);
+    static Map<ShapeId, Location> collectLocations(ValidatedResult<Model> model) {
+        if (model.getResult().isPresent()) {
+            ShapeLocationCollector collector = new FileCachingCollector();
+            return collector.collectDefinitionLocations(model.getResult().get());
+        }
+        return new HashMap<>();
+    }
+
+    public ValidatedResult<Model> getModel() {
```
JordonPhillips, line 275: public methods should have docs

milesziemer:
Done in https://github.com/smithy-lang/smithy-language-server/pull/146

PR #133, opened 2023-11-24
Avoid substring if not suffixed

### LINKED ISSUE #132: Adding the `@output` or `@input` trait on non-inlined struct can break lsp
As far as I can tell, this is a valid model:

```
$version: "2"

namespace test

service Service {
    version: "1"
    operations: [HelloWorld]
}

operation HelloWorld {
    input: Other
}

@input
structure Other {
    @required
    value: String
}
```

But it will crash the lsp because it fails while collecting locations. This happens because the input structure `test#Other` is not suffixed by `Input`. The [code here](https://github.com/smithy-lang/smithy-language-server/blob/bb6b3cc72e40e2d2e884fa753a9e4f2f59f516f9/src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java#L243) assumes it is and does some string manipulation that is wrong.

I think the original code is written with this assumption because inlined input/output shape will be annotated with `@intput`/`@output`, respectively, while also being suffixed. But this is not a hard requirement. The model above is valid and while it emits a warning because `Other` does not start with `HelloWorld`, it is still valid.

daddykotex:
*Issue #, if available:*

Fixes #132 

*Description of changes:*

Avoid the string manipulation if the shape name is not suffixed by input or ouput.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
+++ b/src/main/java/software/amazon/smithy/lsp/ext/FileCachingCollector.java
@@ -240,7 +240,10 @@ final class FileCachingCollector implements ShapeLocationCollector {
             String suffix = getOperationInputOrOutputSuffix(shape, preamble);
             String shapeName = shape.getId().getName();
 
-            String matchingOperationName = shapeName.substring(0, shapeName.length() - suffix.length());
+            String matchingOperationName =
+                shapeName.endsWith(suffix)
+                    ? shapeName.substring(0, shapeName.length() - suffix.length())
+                    : shapeName;
             ShapeId matchingOperationId = ShapeId.fromParts(shape.getId().getNamespace(), matchingOperationName);
 
             return model.shapes(OperationShape.class)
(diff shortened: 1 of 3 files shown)


srchase:
@daddykotex, would you mind adding a couple of structures with @input and @output traits at the end of [this test model](https://github.com/smithy-lang/smithy-language-server/blob/main/src/test/resources/software/amazon/smithy/lsp/ext/models/v2/main.smithy) along with an assertion [here](https://github.com/smithy-lang/smithy-language-server/blob/bb6b3cc72e40e2d2e884fa753a9e4f2f59f516f9/src/test/java/software/amazon/smithy/lsp/ext/SmithyProjectTest.java#L419)?

daddykotex:
Details of the repro in the commit msg: 

https://github.com/smithy-lang/smithy-language-server/pull/133/commits/425d229e65f6a72f7d51d0dec38abf2eccd17c61

srchase (approved):

PR #135, opened 2023-12-13
Prevent the automated workflow from running in fork

daddykotex:
*Issue #, if available:*

No issue available, but when forking this workflow we end up with a scheduled workflow that runs but w/o the token. 

*Description of changes:*

Make the job conditional on whether or not it is executing from _this_ repository.

### DIFF
--- a/.github/workflows/update-smithy-version.yml
+++ b/.github/workflows/update-smithy-version.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   get-version:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


kstich (approved):
Looks like the only way to validate this is to merge, run the workflow manually, and revert if it skips for some reason.

kstich:
Action run failed here, pointing at the string contents. It looks like the GH docs use single quotes (`'`) instead of doubles (`"`),

PR #139, opened 2024-01-18
Use smithy official formatter

daddykotex:
*Issue #, if available:*

N/A

*Description of changes:*

We've implemented support for formatting via the [formatter in smithy-translate](https://github.com/disneystreaming/smithy-translate/tree/main/modules/formatter) but now that the official formatter is out, there is no reason not to use it.

In my own tests, the official formatter produce different results but it is more consistent. Our modeling of the AST is wrong when it comes to whitespaces and as such we produce poorly formatted files when comments are extensively used. For example, [this recent issue](https://github.com/disneystreaming/smithy-translate/issues/226#issuecomment-1898543615) highlight a simple snippet that is poorly formatted. The official formatter does the correct thing here.

I'd be happy to add some tests if you like this PR. Feel free to push changes to this PR if you want to do it.


milesziemer:
Done in https://github.com/smithy-lang/smithy-language-server/pull/146

PR #160, opened 2024-08-09
Add support for multiple workspaces

milesziemer:
Previously, the language server only knew about a single workspace root, so if your editor was using [WorkspaceFolders](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders) the server would just pick the first one, and not load any others. This commit allows the server to load multiple workspaces. The primary challenge was handling state changes to individual workspaces independently. We use client-side file watchers and the `didChangeWatchedFiles` notification to make sure projects are up to date with new and deleted Smithy files, and any changes to build files (i.e. smithy-build.json). `didChangeWatchedFiles` sends a flat list of file events - not partitioned by workspace - so we have to figure out which projects each change applies to. This is done by creating a `PathMatcher` for each project's smithy files and build files, then matching on each file event. This way, we can apply changes to each individual project, rather than reloading everything. Selectors were also updated to select from all available projects.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
@@ -161,7 +161,7 @@ public class SmithyLanguageServer implements
     }
 
     Project getFirstProject() {
-        return projects.nonDetachedProjects().values().stream().findFirst().orElse(null);
+        return projects.attachedProjects().values().stream().findFirst().orElse(null);
     }
 
     ProjectManager getProjects() {
@@ -174,7 +174,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public void connect(LanguageClient client) {
-        LOGGER.info("Connect");
+        LOGGER.finest("Connect");
         this.client = new SmithyLanguageClient(client);
         String message = "smithy-language-server";
         try {
@@ -189,7 +189,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
-        LOGGER.info("Initialize");
+        LOGGER.finest("Initialize");
 
         // TODO: Use this to manage shutdown if the parent process exits, after upgrading jdk
         // Optional.ofNullable(params.getProcessId())
@@ -238,12 +238,12 @@ public class SmithyLanguageServer implements
             }
         }
 
-        LOGGER.info("Done initialize");
+        LOGGER.finest("Done initialize");
         return completedFuture(new InitializeResult(CAPABILITIES));
     }
 
     private void tryInitProject(String name, Path root) {
-        LOGGER.info("Initializing project at " + root);
+        LOGGER.finest("Initializing project at " + root);
         lifecycleManager.cancelAllTasks();
         Result<Project, List<Exception>> loadResult = ProjectLoader.load(
                 root, projects, lifecycleManager.managedDocuments());
@@ -251,7 +251,7 @@ public class SmithyLanguageServer implements
             Project updatedProject = loadResult.unwrap();
             resolveDetachedProjects(this.projects.getProjectByName(name), updatedProject);
             this.projects.updateProjectByName(name, updatedProject);
-            LOGGER.info("Initialized project at " + root);
+            LOGGER.finest("Initialized project at " + root);
         } else {
             LOGGER.severe("Init project failed");
             // TODO: Maybe we just start with this anyways by default, and then add to it
@@ -308,7 +308,7 @@ public class SmithyLanguageServer implements
 
     private CompletableFuture<Void> registerSmithyFileWatchers() {
         List<Registration> registrations = FileWatcherRegistrationHandler.getSmithyFileWatcherRegistrations(
-                projects.nonDetachedProjects().values());
+                projects.attachedProjects().values());
         return client.registerCapability(new RegistrationParams(registrations));
     }
 
@@ -320,7 +320,7 @@ public class SmithyLanguageServer implements
     @Override
     public void initialized(InitializedParams params) {
         List<Registration> registrations = FileWatcherRegistrationHandler.getBuildFileWatcherRegistrations(
-                projects.nonDetachedProjects().values());
+                projects.attachedProjects().values());
         client.registerCapability(new RegistrationParams(registrations));
         registerSmithyFileWatchers();
     }
@@ -348,7 +348,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public CompletableFuture<String> jarFileContents(TextDocumentIdentifier textDocumentIdentifier) {
-        LOGGER.info("JarFileContents");
+        LOGGER.finest("JarFileContents");
         String uri = textDocumentIdentifier.getUri();
         Project project = projects.getProject(uri);
         Document document = project.getDocument(uri);
@@ -362,7 +362,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public CompletableFuture<List<? extends Location>> selectorCommand(SelectorParams selectorParams) {
-        LOGGER.info("SelectorCommand");
+        LOGGER.finest("SelectorCommand");
         Selector selector;
         try {
             selector = Selector.parse(selectorParams.getExpression());
@@ -374,7 +374,7 @@ public class SmithyLanguageServer implements
 
         // Select from all available projects
         Collection<Project> detached = projects.detachedProjects().values();
-        Collection<Project> nonDetached = projects.nonDetachedProjects().values();
+        Collection<Project> nonDetached = projects.attachedProjects().values();
 
         return completedFuture(Stream.concat(detached.stream(), nonDetached.stream())
                 .flatMap(project -> project.modelResult().getResult().stream())
@@ -389,7 +389,7 @@ public class SmithyLanguageServer implements
     @Override
     public CompletableFuture<ServerStatus> serverStatus() {
         List<OpenProject> openProjects = new ArrayList<>();
-        for (Project project : projects.nonDetachedProjects().values()) {
+        for (Project project : projects.attachedProjects().values()) {
             openProjects.add(new OpenProject(
                     LspAdapter.toUri(project.root().toString()),
                     project.smithyFiles().keySet().stream()
@@ -410,7 +410,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
-        LOGGER.info("DidChangeWatchedFiles");
+        LOGGER.finest("DidChangeWatchedFiles");
         // Smithy files were added or deleted to watched sources/imports (specified by smithy-build.json),
         // or the smithy-build.json itself was changed
 
@@ -442,7 +442,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
-        LOGGER.info("DidChangeWorkspaceFolders");
+        LOGGER.finest("DidChangeWorkspaceFolders");
 
         Either<String, Integer> progressToken = Either.forLeft(UUID.randomUUID().toString());
         try {
@@ -487,7 +487,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public void didChange(DidChangeTextDocumentParams params) {
-        LOGGER.info("DidChange");
+        LOGGER.finest("DidChange");
 
         if (params.getContentChanges().isEmpty()) {
             LOGGER.info("Received empty DidChange");
@@ -530,7 +530,7 @@ public class SmithyLanguageServer implements
 
     @Override
     public void didOpen(DidOpenTextDocumentParams params) {
-        LOGGER.info("DidOpen");
+        LOGGER.finest("DidOpen");
 
         String uri = params.getTextDocument().getUri();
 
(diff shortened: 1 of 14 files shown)


src/test/java/software/amazon/smithy/lsp/RequestBuilders.java
```diff
@@ -250,4 +256,24 @@ public DidChangeWatchedFilesParams build() {
             return new DidChangeWatchedFilesParams(changes);
         }
     }
+
+    public static final class DidChangeWorkspaceFolders {
+        final List<WorkspaceFolder> added = new ArrayList<>();
```
hpmellema, line 261: Maybe I just overlooked, but Im not seeing where these are used as package private?

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -378,29 +372,31 @@ public CompletableFuture<List<? extends Location>> selectorCommand(SelectorParam
...
-                projects.mainProject().smithyFiles().keySet().stream()
-                        .map(LspAdapter::toUri)
-                        .collect(Collectors.toList()),
-                false);
-
         List<OpenProject> openProjects = new ArrayList<>();
-        openProjects.add(openProject);
+        for (Project project : projects.nonDetachedProjects().values()) {
+            openProjects.add(new OpenProject(
+                    LspAdapter.toUri(project.root().toString()),
```
hpmellema, line 394: nit: if you do the to-string conversion any other places maybe add an overloaded method on the adapter

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -417,46 +413,72 @@ public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
...
-                    }
-                }
+
+        Map<String, ProjectChanges> changesByProject = projects.computeProjectChanges(params.getChanges());
+
+        changesByProject.forEach((projectName, projectChanges) -> {
+            Project project = projects.getProjectByName(projectName);
+            if (projectChanges.hasChangedBuildFiles()) {
+                client.info("Build files changed, reloading project");
+                // TODO: Handle more granular updates to build files.
```
hpmellema, line 423: Q: Could you provide me an example of what more granular updates you are thinking here?

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -417,46 +413,72 @@ public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
...
-                        changedBuildFiles = true;
-                        break;
-                    }
-                }
+
+        Map<String, ProjectChanges> changesByProject = projects.computeProjectChanges(params.getChanges());
+
+        changesByProject.forEach((projectName, projectChanges) -> {
+            Project project = projects.getProjectByName(projectName);
+            if (projectChanges.hasChangedBuildFiles()) {
```
hpmellema, line 421: I think we should short circuit this if else check for cases with no changes with `if (!projectChanges.hasChangedBuildFiles() || projectChanges.hasChangedSmithyFiles()) { return; }` unless the `ProjectChanges` can never be empty in that way?

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -417,46 +413,72 @@ public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
...
+
+        // TODO: Update watchers based on specific changes
+        unregisterSmithyFileWatchers().thenRun(this::registerSmithyFileWatchers);
+
+        sendFileDiagnosticsForManagedDocuments();
+    }
+
+    @Override
+    public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
+        LOGGER.info("DidChangeWorkspaceFolders");
```
hpmellema, line 445: Should this be debug? Seems like it is a bit noisy for info

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -417,46 +413,72 @@ public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
...
+
+        Either<String, Integer> progressToken = Either.forLeft(UUID.randomUUID().toString());
+        try {
+            client.createProgress(new WorkDoneProgressCreateParams(progressToken)).get();
+        } catch (ExecutionException | InterruptedException e) {
```
hpmellema, line 450: Why do we block on creating this progress token?

src/main/java/software/amazon/smithy/lsp/project/ProjectChanges.java
```diff
@@ -0,0 +1,35 @@
...
+public record ProjectChanges(
+        Set<String> changedBuildFileUris,
+        Set<String> createdSmithyFileUris,
+        Set<String> deletedSmithyFileUris
+) {
+    /**
+     * @return Whether there are any changed build files
+     */
+    public boolean hasChangedBuildFiles() {
+        return !changedBuildFileUris.isEmpty();
```
hpmellema, line 26: Can all 3 ever be empty?

src/main/java/software/amazon/smithy/lsp/project/ProjectFilePatterns.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * @param project The project to get the watch pattern for
+     * @return A glob pattern used to watch build files in the given project
+     */
+    public static String getBuildFilesWatchPattern(Project project) {
+        Path root = project.root();
+        String buildJsonPattern = escapeBackslashes(root.resolve(ProjectConfigLoader.SMITHY_BUILD).toString());
+        String projectJsonPattern = escapeBackslashes(root.resolve(ProjectConfigLoader.SMITHY_PROJECT).toString());
+
+        List<String> patterns = new ArrayList<>();
```
hpmellema, line 55: nit: Is it possible to right-size this array?

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
```diff
@@ -16,39 +22,51 @@
...
      *  corresponds to the key in the map.
      */
     public Map<String, Project> detachedProjects() {
         return detached;
     }
 
+    /**
+     * @return A map of project names to projects tracked by the server
+     */
+    public Map<String, Project> nonDetachedProjects() {
```
hpmellema, line 66: nit: Why not "attached" projects?

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
```diff
@@ -57,12 +75,13 @@ public Project getProject(String uri) {
...
-            //  be tracking any files that aren't attached to a project. But for testing, this
-            //  is useful to ensure that fact.
+        }  else  {
+            for (Project project : projects.values()) {
+                if (project.smithyFiles().containsKey(path)) {
+                    return project;
+                }
+            }
+
             return null;
```
hpmellema, line 85: Should this log an error or warning of some sort that the uri could not be found?

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
```diff
@@ -126,4 +155,51 @@ public Document getDocument(String uri) {
...
+        Map<String, PathMatcher> projectBuildFileMatchers = new HashMap<>(nonDetachedProjects().size());
+
+        Map<String, ProjectChanges> changes = new HashMap<>(nonDetachedProjects().size());
+
+        nonDetachedProjects().forEach((projectName, project) -> {
+            projectSmithyFileMatchers.put(projectName, ProjectFilePatterns.getSmithyFilesPathMatcher(project));
+            projectBuildFileMatchers.put(projectName, ProjectFilePatterns.getBuildFilesPathMatcher(project));
+
+            // Need these to be hash sets so they are mutable
+            changes.put(projectName, new ProjectChanges(new HashSet<>(), new HashSet<>(), new HashSet<>()));
```
hpmellema, line 178: Nit: You can make an empty constructor for the record that does this.

Also, do we care at all about the order of project changes?

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
```diff
@@ -126,4 +155,51 @@ public Document getDocument(String uri) {
...
+        });
+
+        for (FileEvent event : events) {
+            String changedUri = event.getUri();
+            Path changedPath = Path.of(LspAdapter.toPath(changedUri));
+            if (changedUri.endsWith(".smithy")) {
+                projectSmithyFileMatchers.forEach((projectName, matcher) -> {
+                    if (matcher.matches(changedPath)) {
+                        if (event.getType() == FileChangeType.Created) {
```
hpmellema, line 187: Do we expect there to be more file changes types? If so a switch might be nice here.

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
```diff
@@ -126,4 +155,51 @@ public Document getDocument(String uri) {
...
+                            changes.get(projectName).createdSmithyFileUris().add(changedUri);
```
hpmellema, line 188: Thought: What if we did "getOrCompute" here? Then we could only create the changes for projects that have them.

src/test/java/software/amazon/smithy/lsp/RequestBuilders.java
hpmellema: it looks to me like these builder members can just be private.

hpmellema (changes requested):

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
milesziemer: Probably all of them should

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
milesziemer: It is async in that it sends a request to the client, but I'm not sure if that was your question. I want this method to be mostly synchronous, since we're mutating the project states

hpmellema: You indirectly answered it. I was mostly wondering if we were OK with the blocking, and it seems like we are.

src/main/java/software/amazon/smithy/lsp/project/ProjectChanges.java
milesziemer: Maybe, if the protocol changes or we change how we compute the changes (i.e. if there's a non-meaningful change to smithy-build.json)

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
milesziemer: Rare, not using `Path` much

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
milesziemer: build files and smithy file changes aren't mutually exclusive, need to fix this conditional

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
milesziemer: Because naming is hard

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
milesziemer: No

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
milesziemer: No, the hashsets are already initialized

milesziemer: Note: if you have lots of projects, this will create a bunch of garbage because most times you aren't getting multiple sets of large updates to a project

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
milesziemer: Actually, looking at this again, when build files are changed it will perform the work that would have been done for added/removed smithy files, so the conditional is correct

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
milesziemer: Decided not to do this, because I like how this way is explicit about needing the sets to be mutable, whereas it isn't clear an empty ctor should init with mutable sets.

src/test/java/software/amazon/smithy/lsp/RequestBuilders.java
milesziemer: Made them all private

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
milesziemer: Made all these trace messages `finest` (there's no debug)

src/main/java/software/amazon/smithy/lsp/project/ProjectManager.java
milesziemer: Added a note

hpmellema (approved):

PR #161, opened 2024-08-11
Use newer java features

milesziemer:
Refactors various things to make use of newer java features, including
record classes, text blocks, and new APIs.

Depends on https://github.com/smithy-lang/smithy-language-server/pull/160


src/main/java/software/amazon/smithy/lsp/document/DocumentId.java
```diff
@@ -13,7 +13,7 @@
  * inaccurate in the sense that the string value it references isn't
  * necessarily a valid identifier, it just looks like an identifier.
  */
-public final class DocumentId {
+public record DocumentId(Type type, CharBuffer idSlice, Range range) {
```
hpmellema, line 16: nit: Param javadocs?

src/main/java/software/amazon/smithy/lsp/document/DocumentParser.java
```diff
@@ -643,33 +644,12 @@ private boolean isShapeType() {
...
-            case "service":
-            case "resource":
-            case "bigDecimal":
-            case "bigInteger":
-                return true;
-            default:
-                return false;
-        }
+        return switch (token.toString()) {
+            case "structure", "operation", "string", "integer", "list", "map", "boolean", "enum", "union", "blob",
```
hpmellema, line 648: Could we change this to a static hashset with a `contains` check?

hpmellema (changes requested):

src/main/java/software/amazon/smithy/lsp/document/DocumentParser.java
milesziemer: Maybe. I wonder how that compares to the lookup in the string pool (which I'm guessing this switch does, in some way). Or if that's relevant

hpmellema (approved):

PR #162, opened 2024-08-26
Replace DocumentParser impl with a real parser

milesziemer:
Depends on #160 and #161 , leaving this in draft until those are merged.

This commit is the first of a few that refactor the 'frontend' of the
server, i.e. the completions, definition, hover, etc. It replaces the
implementation of DocumentParser, which was used to compute specific
syntactic information about a Smithy file, with an actual parser that
will be used in future commits to overhaul language features like
completions.

For now, this is mostly a 1:1 swap - DocumentParser now
does no parsing, but instead uses the results of the actual parser to
compute the same info it used to. The only other real change to
functionality is that DocumentParser no longer depends on the result of
loading the model, and SmithyFiles are re-parsed synchronously on
changes.

The javadoc for the new `Syntax` class contains more detailed
information on how the parser works, and why it works that way, but some
of it may not make complete sense with the current code in the language
server. The parser was built in tandem with the upgrades to completions,
and was specifically designed to enhance such features. Those upgrades
deserve their own commit, so I didn't include them here.


milesziemer:
Closing in favor of https://github.com/smithy-lang/smithy-language-server/pull/166

PR #167, opened 2024-09-25
Support projects in subdirectories

milesziemer:
This commit makes the language server work with Smithy projects in subdirectories of the folder (or folders, in the case of e.g. vscode workspace folders) open in the editor. I previously added support for multiple _workspace folders_ (an LSP concept), but I assumed only one _project_ (Smithy LS concept) per workspace folder. So this commit fixes that mixing, allowing many projects per workspace folder.

Now, the language server will search through subdirectories of all workspace folders (by default, one workspace folder is open in the client) to find projects. Changes to build files, i.e. smithy-build.json, .smithy-project.json, are now tracked at the workspace level, so you can add a new project to an existing workspace.

I also did _some_ cleanup of the project/workspace synchronization code, and moved some things around. A note on some tests: I'm using a `Files.createTempDirectory`, and adding the `TestWorkspace` as a subdir. In a follow-up commit, I will be changing `TestWorkspace` to be something like `TestProject`, which is more accurate. I didn't include it here to avoid a bunch of noise.

Also:

I fixed a bug in https://github.com/smithy-lang/smithy-language-server/pull/160 that caused a deadlock. 

*Issue #, if available:*

- https://github.com/smithy-lang/smithy-vscode/issues/95
- https://github.com/smithy-lang/smithy-vscode/issues/92


cgeduhn:
We Are looking forward for this 🎉

hpmellema (changes requested):

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -443,40 +468,54 @@ public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
...
+        for (var entry : projects.attachedProjects().entrySet()) {
+            if (entry.getValue().root().startsWith(workspaceRoot)) {
+                projectsToRemove.add(entry.getKey());
+            }
         }
 
-        unregisterSmithyFileWatchers().thenRun(this::registerSmithyFileWatchers);
-        sendFileDiagnosticsForManagedDocuments();
+        for (String projectName : projectsToRemove) {
+            removeProjectAndResolveDetached(projectName);
```
hpmellema, line 511: do we need to do this in a separate loop from the above? Could we just remove as we go?

milesziemer: There's a comment above the above loop:
```
        // Have to do the removal separately, so we don't modify project.attachedProjects()
        // while iterating through it
```
Removing modifies the attachedProjects map, which is what backs the entrySet being iterated over

src/main/java/software/amazon/smithy/lsp/FilePatterns.java
```diff
@@ -38,45 +40,71 @@ public static List<String> getSmithyFileWatchPatterns(Project project) {
...
+
+    // Patterns for the workspace need to match on all build files in all subdirectories,
+    // whereas patterns for projects only look at the top level (because project locations
+    // are defined by the presence of these build files).
+    private static String getBuildFilesPattern(Path root, boolean isWorkspacePattern) {
+        List<String> patterns = new ArrayList<>(BUILD_FILE_COUNT);
+        patterns.add(ProjectConfigLoader.SMITHY_BUILD);
+        patterns.add(ProjectConfigLoader.SMITHY_PROJECT);
+        for (String buildExt : ProjectConfigLoader.SMITHY_BUILD_EXTS) {
+            Path extPath = Path.of(buildExt); // buildExt may have file separators
```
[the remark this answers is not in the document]
milesziemer: Done

hpmellema (approved):

PR #168, opened 2024-10-03
Track build file changes

milesziemer:
This change makes the server register for document lifecycle events like
didChange for build files. Previous behavior was to register for file
system change events for build files, but eventually we want to give
diagnostics, completions, etc. for build files as you type, so we need
to track all changes. This change keeps existing behavior from the
user's perspective, as we still only reload projects on save, so it
serves mostly as a refactor.

I also extracted the state managed by SmithyLanguageServer to its own
class, ServerState, and flattened ProjectManager into that class, so we
can manage the state of the server in a more centralized location.

Depends on https://github.com/smithy-lang/smithy-language-server/pull/167


src/main/java/software/amazon/smithy/lsp/project/ProjectConfig.java
```diff
@@ -90,14 +93,13 @@ static final class Builder {
         String outputDirectory;
         final List<ProjectDependency> dependencies = new ArrayList<>();
         MavenConfig mavenConfig;
-        final List<Path> loadedConfigPaths = new ArrayList<>();
+        final Map<String, BuildFile> buildFiles = new HashMap<>();
```
hpmellema, line 96: nit: can be private.

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -718,13 +680,17 @@ public CompletableFuture<CompletionItem> resolveCompletionItem(CompletionItem un
...
         String uri = params.getTextDocument().getUri();
-        if (!projects.isTracked(uri)) {
+        ProjectAndFile projectAndFile = state.findProjectAndFile(uri);
+        if (projectAndFile == null) {
             client.unknownFileError(uri, "definition");
-            return completedFuture(Either.forLeft(Collections.emptyList()));
+            return completedFuture(null);
+        }
+
+        if (!(projectAndFile.file() instanceof SmithyFile smithyFile)) {
```
hpmellema, line 689: Nit: 3 repetitions of this logic. I wonder if we could consolidate

src/main/java/software/amazon/smithy/lsp/ServerState.java
```diff
@@ -0,0 +1,236 @@
...
+        return getManagedDocument(uri);
+    }
+
+    ProjectAndFile findProjectAndFile(String uri) {
+        String path = LspAdapter.toPath(uri);
+        // TODO: From isDetached
+        for (Project project : attachedProjects.values()) {
+            ProjectFile projectFile = project.getProjectFile(path);
+            if (projectFile != null) {
+                detachedProjects.remove(uri);
```
hpmellema, line 94: Im a little confused why we remove it from detached projects here? I would have thought that would get checked correctly in the isDetached below?

src/main/java/software/amazon/smithy/lsp/ServerState.java
```diff
@@ -0,0 +1,236 @@
...
+        return null;
+    }
+
+    boolean isDetached(String uri) {
+        // We might be in a state where a file was added to a tracked project,
+        // but was opened before the project loaded. This would result in it
+        // being placed in a detachedProjects project. Removing it here is basically
+        // like removing it lazily, although it does feel a little hacky.
+        String path = LspAdapter.toPath(uri);
+        if (detachedProjects.containsKey(uri)) {
```
hpmellema, line 119: nit: I think maybe we could invert this to return early.

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
milesziemer: In the future, there will actually be different paths for BuildFiles. We will have to refactor it then anyways, but with the knowledge of how it needs to end up working

src/main/java/software/amazon/smithy/lsp/ServerState.java
milesziemer: Yea it does, but I'm not calling that method here. I think I meant to end up removing `isDetached`, but it ended up being too useful. I did some refactoring and pulled the removal logic into the new `findAttachedAndRemoveDetached`

src/main/java/software/amazon/smithy/lsp/ServerState.java
milesziemer: I refactored this a bit, and now I like it within the if since it's more clear (to me at least) that we need detachedProjects to contain the uri.

hpmellema (approved):

PR #184, opened 2025-01-04
 Add ShapeId as detail in completions 

kubukoz:
It can be useful when a given name shows up in many namespaces and you want to choose which one gets imported.

image


kubukoz:
I should've remembered about #166 first... but given this is a one-liner (plus tests), perhaps we can have this without waiting for the other PR? (which is massive, in a good way :) )

yasmewad (approved):
LGTM.

milesziemer:
https://github.com/smithy-lang/smithy-language-server/pull/166 finally got merged, so will close this one

kubukoz:
Looking forward to that release :)

PR #185, opened 2025-01-14
moved initializeParams to serverOptions class

divmeh-aws:
*Issue #, if available:*

https://github.com/smithy-lang/smithy-language-server/issues/171
Bullet point 1 of the issue is addressed here.

*Description of changes:*

- New class ServerOptions validates and saves config options sent by the client. This makes it neatly extensible when we add new options down the line. 


src/main/java/software/amazon/smithy/lsp/ServerOptions.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Creates a ServerOptions instance from the initialization options provided by the client.
+ * Parses and validates configuration settings from the initialization parameters.
+ *
+ * @param initializationOptions The raw initialization options object from the client,
+ *                             expected to be a JsonObject containing server configurations
+ * @param client The language client used for logging configuration status and errors
+ * @return A new {@code ServerOptions} instance with parsed configuration values
+ **/
+    public static ServerOptions fromInitializeParams(Object initializationOptions, SmithyLanguageClient client) {
```
milesziemer, line 54: Now that I see this, I don't think I like my original suggestion of doing it this way, since its not clear that `initializationOptions` comes from InitializeParams. Maybe just pass in the whole `InitializeParams`.

src/main/java/software/amazon/smithy/lsp/ServerOptions.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+            if (jsonObject.has("onlyReloadOnSave")) {
+                builder.setOnlyReloadOnSave(jsonObject.get("onlyReloadOnSave").getAsBoolean());
+                client.info("Configured only reload on save: " + builder.onlyReloadOnSave);
+            }
+        }
+        return builder.build();
+    }
+
+    public static class Builder {
```
milesziemer, line 77: Suggestion: You can make `Builder` final, since we're not subclassing it
```suggestion
    public static final class Builder {
```

src/main/java/software/amazon/smithy/lsp/ServerOptions.java
```diff
@@ -0,0 +1,95 @@
...
+
+//import javax.print.attribute.standard.Severity;
+import com.google.gson.JsonObject;
+import java.util.Arrays;
+import java.util.Optional;
+import software.amazon.smithy.model.validation.Severity;
+
+public final class ServerOptions {
+    private Severity minimumSeverity = Severity.WARNING;
+    private boolean onlyReloadOnSave = false;
```
milesziemer, line 26: Suggestion: You can make these final, since at least right now we aren't going to change them after initialization (supporting that would require adding some synchronization anyway). This also lets `Builder` be the source-of-truth for the default values of each option.
```suggestion
    private final Severity minimumSeverity;
    private final boolean onlyReloadOnSave;
```

src/main/java/software/amazon/smithy/lsp/ServerOptions.java
```diff
@@ -0,0 +1,95 @@
...
+    private ServerOptions(Builder builder) {
+        this.minimumSeverity = builder.minimumSeverity;
+        this.onlyReloadOnSave = builder.onlyReloadOnSave;
+    }
+
+    public Severity getMinimumSeverity() {
+        return this.minimumSeverity;
+    }
+
+    public Boolean getOnlyReloadOnSave() {
```
milesziemer, line 37:
```suggestion
    public boolean getOnlyReloadOnSave() {
```

milesziemer (changes requested):

divmeh-aws (commented):
addressed requested changes.

milesziemer (approved):

PR #187, opened 2025-01-17
Cleanup serverstate and project

milesziemer:
Various refactoring commits that clean up ServerState and Project. Exposing the fields of ServerState made it super annoying to refactor, so I made ServerState a class and hid everything. Then, I was able to make ServerState only store one collection of projects, and handle its own mutations. Similar story for Project. 


src/main/java/software/amazon/smithy/lsp/ManagedFiles.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package software.amazon.smithy.lsp;
+
+import software.amazon.smithy.lsp.document.Document;
+
+/**
+ * Provides access to {@link Document}s managed by the server.
+ *
+ * <p>A document is _managed_ if its state is controlled by the lifecycle methods
```
yefrig, line 13: nit: if <p> is a separator then maybe keep it on its own line

src/main/java/software/amazon/smithy/lsp/ServerState.java
```diff
@@ -10,57 +10,71 @@
...
+     * @return All projects tracked by the server.
      */
+    public Collection<Project> getAllProjects() {
+        return projects.values();
+    }
+
+    /**
+     * @return All files managed by the server, including their projects.
+     */
+    public Collection<ProjectAndFile> getAllManaged() {
```
yefrig, line 62: if this is public, you might want to add it to the ManagedFiles interface. just a thought

yefrig (approved):

src/main/java/software/amazon/smithy/lsp/ServerState.java
milesziemer: Might make sense, or something similar. Right now ManagedFiles doesn't really know about Project, so we will have to see

Issue #189, opened 2025-01-30
Implement folding range


milesziemer:
Some editors have client-side support for folding ranges based on treesitter or textmate grammar, but we should support folding range server-side. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange.

milesziemer:
https://github.com/smithy-lang/smithy-language-server/pull/190

PR #190, opened 2025-02-05
Added FoldingRange Feature to LSP

joewyz:
*Issue #, if available:*
Issue #189 
*Description of changes:*
Added FoldingRange Feature and unit tests to LSP.
Added position setting for obj's kvps field.
Added end position setting for kvp node.


src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
```diff
@@ -0,0 +1,198 @@
...
+                case Syntax.Statement.Metadata metadata -> processFoldingRangeForNode(metadata.value(), foldingRanges);
+
+                case Syntax.Statement.Block blk -> processFoldingRangeForBlock(blk, foldingRanges);
+
+                case Syntax.Statement.NodeMemberDef nodeMember ->
+                        processFoldingRangeForNode(nodeMember.value(), foldingRanges);
+
+                // For most of the current statement types just create the folding range based on
+                // its own range
+                default -> addFoldingRange(statement.start(), statement.end(), foldingRanges);
```
milesziemer, line 103: I wonder if in practice this ever ends up actually adding a range, since most statements (except the ones already checked in this switch) are single line

src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
```diff
@@ -0,0 +1,198 @@
...
+                if (arr.elements().isEmpty()) {
+                    return;
+                }
+
+                addFoldingRange(arr.start(), arr.end(), foldingRanges);
+
+                arr.elements().forEach(element -> processFoldingRangeForNode(element, foldingRanges));
+            }
+
+            default -> addFoldingRange(node.start(), node.end(), foldingRanges);
```
milesziemer, line 194: Similar to my previous comment, I wonder if this ends up creating a range. Maybe if its a string block

src/main/java/software/amazon/smithy/lsp/syntax/Parser.java
```diff
@@ -813,15 +815,16 @@ private void enumMember(Syntax.Statement.Block parent) {
             Syntax.Ident name = ident();
             var enumMemberDef = new Syntax.Statement.EnumMemberDef(parent, name);
             enumMemberDef.start = start;
+            setEnd(enumMemberDef);
```
milesziemer, line 818: What's the difference between setting the end two different times here, vs how it was before, just setting it once after the if?

milesziemer (changes requested):

src/main/java/software/amazon/smithy/lsp/syntax/Parser.java
milesziemer: Nevermind, I get it - because the whitespace gets consumed.

src/test/java/software/amazon/smithy/lsp/language/FoldingRangeHandlerTest.java
```diff
@@ -0,0 +1,463 @@
...
+    }
+
+
+
+    @Test
+    public void foldingRangeForMultipleAndNestedTraits() {
+        TextWithPositions model = TextWithPositions.from("""
+                @restJson1%
+                @title("")
+                @cors()
```
milesziemer, line 201: For cases like this, need to change Parser#traitNode() to not skip the closing ')', since that causes Parser#traitApplication() to eat all the whitespace after the closing ')', setting the end of the TraitApplication to be the start of the next statement

src/main/java/software/amazon/smithy/lsp/syntax/Parser.java
milesziemer: Can just add a comment here to explain why

src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
```diff
@@ -0,0 +1,198 @@
...
+     * Main public handle function in the handler class
+     *
+     * @return A list of FoldingRange
+     */
+    public List<FoldingRange> handle() {
+        return generateFoldingRanges();
+
+    }
+
+    /**
```
milesziemer, line 31: Nits: No need for docs on private methods

src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
```diff
@@ -0,0 +1,198 @@
...
+        return range != null && range.getEnd().getLine() - range.getStart().getLine() >= 1;
+    }
+
+    /**
+     * Create a fold range object with given range and add it to the foldingRanges list
+     *
+     * @param range         The Range object held the start line and end line info for the folding range
+     * @param foldingRanges The list holds folding ranges for current document
+     */
+    private void addFoldingRange(Range range, List<FoldingRange> foldingRanges) {
```
milesziemer, line 47:
```suggestion
    private void addFoldingRange(List<FoldingRange> foldingRanges, Range range) {
```
On all these methods, can put the ranges list first, so if we change any of these we don't need to change much on the call-sites, since ranges will always be first

src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Check if the given range can be folded
+     *
+     * @param range The Range object held the start line and end line info for the folding range
+     * @return If the given Range can be folded (end line - start line >= 1)
+     */
+    private boolean isFoldable(Range range) {
+        return range != null && range.getEnd().getLine() - range.getStart().getLine() >= 1;
```
milesziemer, line 38:
```suggestion
        return range != null && range.getEnd().getLine() > range.getStart().getLine();
```

milesziemer, line 38:
```suggestion
        return range != null && range.getEnd().getLine() > range.getStart().getLine();
```

src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
milesziemer, line 47: Another suggestion - you can avoid allocating the intermediate `Range` by just using `Document#lineOfIndex` for the start and end. That method can return -1 if the line doesn't exist (basically equivalent to the range null check).

src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
```diff
@@ -5,194 +5,143 @@
...
     }
 
-    /**
-     * Use the import range in DocumentImport class to create the folding range directly
-     *
-     * @param documentImports The document import object that held the import range
-     * @param foldingRanges   The list holds folding ranges for current document
-     */
-    private void addFoldingRangeForImports(DocumentImports documentImports, List<FoldingRange> foldingRanges) {
+    private void addFoldingRangeForImports(List<FoldingRange> foldingRanges, DocumentImports documentImports) {
```
milesziemer, line 42: Nit: Don't need to pass `documentImports` since it's already a class member

src/main/java/software/amazon/smithy/lsp/language/FoldingRangeHandler.java
```diff
@@ -5,194 +5,143 @@
...
         if (node == null) {
             return;
         }
 
-        Range range = document.rangeBetween(node.start(), node.end());
+        int startLine = document.lineOfIndex(node.start());
+        int endLine = document.lineOfIndex(node.end());
 
-        if (!isFoldable(range)) {
+        if (!isFoldable(startLine, endLine)) {
```
milesziemer, line 122: Is this check redundant? Won't it be checked within the switch?

milesziemer (approved):

PR #193, opened 2025-02-12
Add completions for for build files

milesziemer:
The server now can provide completions for smithy-build.json and .smithy-project.json. The implementation works roughly the same as other node-like completions, using a new builtins model that specifies the structure of the build files. I also had to override the completion item mapper to make sure it wrapped object keys in strings, which is necessary in json.

### DIFF
--- a/src/main/resources/software/amazon/smithy/lsp/language/build.smithy
+++ b/src/main/resources/software/amazon/smithy/lsp/language/build.smithy
@@ -1,0 +1,90 @@
+$version: "2.0"
+
+namespace smithy.lang.server
+
+structure SmithyProjectJson {
+    sources: Strings
+    imports: Strings
+    outputDirectory: String
+    dependencies: ProjectDependencies
+}
+
+list ProjectDependencies {
+    member: ProjectDependency
+}
+
+structure ProjectDependency {
+    name: String
+
+    @required
+    path: String
+}
+
+structure SmithyBuildJson {
+    @required
+    version: SmithyBuildVersion
+
+    outputDirectory: String
+    sources: Strings
+    imports: Strings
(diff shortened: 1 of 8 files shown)


src/main/resources/software/amazon/smithy/lsp/language/build.smithy
```diff
@@ -0,0 +1,90 @@
...
+    dependencies: Strings
+    repositories: MavenRepositories
+}
+
+list MavenRepositories {
+    member: MavenRepository
+}
+
+structure MavenRepository {
+    @required
```
yefrig, line 80: How are constraints used by the LS? Could we possibly add more constraints to save energy doing validations like urls etc

yefrig (approved):
LGTM. 

Tested in neovim v0.11. All completions work fine.

It might be nice to include a text edit with a comma if completing inside a list where the completion is not the last member of the list. But might be out of scope.

joewyz (approved):

milesziemer:
> LGTM.
> 
> Tested in neovim v0.11. All completions work fine.
> 
> It might be nice to include a text edit with a comma if completing inside a list where the completion is not the last member of the list. But might be out of scope.

Nice. Yea that would be cool, maybe a future improvement

src/main/resources/software/amazon/smithy/lsp/language/build.smithy
milesziemer: They really aren't right now. Originally I intended to use them for validation, but I scrapped that idea to use the validation errors from smithy-build. However, with trait values, required members get autopopulated in completions, so maybe a future update could do that for node values too.

PR #199, opened 2025-02-14
Fix didopen for build files

milesziemer:
https://github.com/smithy-lang/smithy-language-server/pull/168 started
tracking build file changes via lifecycle methods, didOpen, etc. But it
didn't make a distinction between what was a build file, and what was a
Smithy file. There are two paths didOpen can take - the first is when
the file being opened is known to be part of a project. In this case,
the file is already tracked by its owning Project, so its basically a
no-op. The second path is when the file does not belong to any project,
in which case we created a "detached" project, which is a project with
no build files and just a single Smithy file. So if you opened a build
file that wasn't known to be part of a Project, the language server
tried to make a detached project containing the build file _as a smithy
file_. This is obviously wrong, but wasn't observable to clients AFAICT
because clients weren't set up to send requests to the server for build
files (specifically, you wouldn't get diagnostics or anything, only for
.smithy files). However, recent commits, including
https://github.com/smithy-lang/smithy-language-server/pull/188, now want
to provide language support for smithy-build.json. In testing these new
commits with local changes to have VSCode send requests for
smithy-build.json, the issue could be observed. Specifically, the issue
happens when you open a new smithy-build.json before we receive the
didChangeWatchedFiles notification that tells us a new build file was
created. didChangeWatchedFiles is the way we actually updated the state
of projects to include new build files, or create new Projects. Since we
can receive didOpen for a build file before didChangeWatchedFiles, we
needed to do something with the build file on didOpen.

This commit addresses the problem by adding a new Project type,
`UNRESOLVED`, which is a project containing a single build file that no
existing projects are aware of. We do this by modifying the didOpen path
when the file isn't known to any project, checking if it is a build file
using a PathMatcher, and if it is, creating an unresolved project for
it. Then, when we load projects following a didChangeWatchedFiles, we
just drop any unresolved projects with the same path as any of the build
files in the newly loaded projects (see ServerState::resolveProjects).

I also made some (upgrades?) to FilePatterns to better handle the
discrepancy between matching behavior of PathMatchers and clients
(see https://github.com/smithy-lang/smithy-language-server/issues/191).
Now there are (private) `*PatternOptions` enums that FilePatterns uses to
configure the pattern for different use cases. For example, the new
FilePatterns::getSmithyFileWatchPathMatchers provides a list of
PathMatchers which should match the same paths as the _watcher_ patterns
we send back to clients, which is useful for testing.

I also fixed an issue where parsing an empty build file would cause an NPE
when trying to map validation events to ranges. Document::rangeBetween 
would return `null` if the document was empty, but I wasn't checking for that
in ToSmithyNode (which creates parse events).

The reason the range is `null` is because Document.lineOfIndex returns
oob for an index of 0 into an empty document. Makes sense, as an empty
document has no lines. I updated a DocumentTest to clarify this
behavior.


src/main/java/software/amazon/smithy/lsp/ServerState.java
```diff
@@ -261,6 +276,24 @@ private void resolveDetachedProjects(Project oldProject, Project updatedProject)
...
+            String uri = LspAdapter.toUri(buildFilePath);
+            projects.remove(uri);
+        }
+    }
+
+    private void removeDetachedProjects(Set<String> smithyFilePaths) {
+        for (String smithyFilePath : smithyFilePaths) {
+            String uri = LspAdapter.toUri(smithyFilePath);
+            projects.remove(uri);
         }
```
joewyz, line 297: These two functions are identical, we can probably just make them to one function

joewyz (approved):

PR #212, opened 2025-03-12
Expand tests on parser to acheive higher coverage and minor fixs

joewyz:
*Description of changes:*
Expand tests on parser to acheive a coverage of 98% in Line and 88% in branch. 
Add fix for Node with leading comma.
Add fix for duplicate statement when invalid member def statement appears.
Add fix for unset end of kvp.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/syntax/Parser.java
+++ b/src/main/java/software/amazon/smithy/lsp/syntax/Parser.java
@@ -55,7 +55,7 @@ final class Parser extends SimpleParser {
                 int start = position();
                 do {
                     skip();
-                } while (!isWs() && !isNodeStructuralBreakpoint() && !eof());
+                } while (!isWs() && !isNodeStructuralBreakpoint() && !eof() && is(','));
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/smithy/lsp/syntax/Parser.java
```diff
@@ -55,7 +55,7 @@ Syntax.Node parseNode() {
                 int start = position();
                 do {
                     skip();
-                } while (!isWs() && !isNodeStructuralBreakpoint() && !eof());
+                } while (!isWs() && !isNodeStructuralBreakpoint() && !eof() && is(','));
```
joewyz, line 58: Not sure if this fix is necessary.

joewyz: If one array start with comma eg. "[,a,b,c]" the first element "a" will be consumed. Adding this will skip the leading comma.

milesziemer (approved):

PR #213, opened 2025-03-26
Implement references and rename

milesziemer:
Adds support for 'textDocument/references', 'textDocument/rename', and
'textDocument/prepareRename'. These all work very similarly, so I
implemented them together.

This implementation only works for root shapes, not members or resource
identifiers/properties (which I call resource fields for brevity).
Originally I intended to make it work for members, but for 'complete'
member support, you really need to support resource fields, otherwise if
you tried to rename a member, it wouldn't rename the resource field it
references, and finding references wouldn't show fully accurate
information. The problem is, the relationship between members and
resource fields can be implicit, and/or determined by a number of
different traits (`resourceIdentifier`, `property`, `references`,
`nestedProperties`). And some of these traits refer to the resource
field by string name. Current KnowledgeIndex implementations don't
support going from an arbitrary member -> field bindings and back. Plus,
we can't reliably use KnowledgeIndicies because they usually operate on
a complete, valid model, whereas the language server wants to operate on
the minimum subset of a possibly invalid model.

I think it would be possible still to support members in the future,
either with some clever code in the language server or enhancements to
the smithy library, but I didn't want to block rename/references for
regular shapes on that.

Some specific implementation details to note:
1. Rename/References throw ResponseErrorException on an unsupported
   position (like a member name), or when the rename would be invalid
   (i.e. renaming a shape from a dependency, or providing an invalid
   shape name). This is what the spec says you should do, and we should
   update other features to do the same.
2. Rename disambiguates shapes that would be conflicting after the rename
   by using an absolute shape id, and removes conflicting use statements.
3. Rename/References both work on references to members of the target
   shape, i.e. `com.foo#Foo$bar`, only if the request position is before
   the `$`, otherwise it assumes you're getting the reference of a
   member.
4. Prepare returns the range of the identifier to be renamed (or throws
   an error as specified in 1.), so if you go to rename `com.foo#Bar`,
   the rename range will be the range of `Bar`.

Other notable changes:
1. Changed `DocumentId` to only have distinct types for root shapes and
   members. The type will be `ROOT` when the cursor is in any part of
   the shape id before a member, and `MEMBER` only when the cursor is
   actually within the member part of the shape id. The rest of the
   other possible id types didn't seem to be useful enough to keep.
2. Added a `line` field to Syntax.Node.Str, which makes it _much_
   more efficient to get the range of a Str or Ident.
3. Fixed an issue in TextWithPositions when there are multiple positions
   on the same line.

Rev: Fix shape search resolution

Fixes an issue with ShapeSearch::findShape where the local namespace was
checked before imports. It was inconsistent with the way the model
loader resolves shapes, so you could end up jumping to the wrong
definition/references, or renaming the wrong shape.

### DIFF
--- a/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
+++ b/src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
@@ -62,11 +62,17 @@ import org.eclipse.lsp4j.InlayHint;
 import org.eclipse.lsp4j.InlayHintParams;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.LocationLink;
+import org.eclipse.lsp4j.PrepareRenameDefaultBehavior;
+import org.eclipse.lsp4j.PrepareRenameParams;
+import org.eclipse.lsp4j.PrepareRenameResult;
 import org.eclipse.lsp4j.ProgressParams;
 import org.eclipse.lsp4j.PublishDiagnosticsParams;
 import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.ReferenceParams;
 import org.eclipse.lsp4j.Registration;
 import org.eclipse.lsp4j.RegistrationParams;
+import org.eclipse.lsp4j.RenameOptions;
+import org.eclipse.lsp4j.RenameParams;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.SetTraceParams;
 import org.eclipse.lsp4j.SymbolInformation;
@@ -82,10 +88,12 @@ import org.eclipse.lsp4j.UnregistrationParams;
 import org.eclipse.lsp4j.WorkDoneProgressBegin;
 import org.eclipse.lsp4j.WorkDoneProgressCancelParams;
 import org.eclipse.lsp4j.WorkDoneProgressEnd;
+import org.eclipse.lsp4j.WorkspaceEdit;
 import org.eclipse.lsp4j.WorkspaceFolder;
 import org.eclipse.lsp4j.WorkspaceFoldersOptions;
 import org.eclipse.lsp4j.WorkspaceServerCapabilities;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.lsp4j.jsonrpc.messages.Either3;
 import org.eclipse.lsp4j.services.LanguageClient;
 import org.eclipse.lsp4j.services.LanguageClientAware;
 import org.eclipse.lsp4j.services.LanguageServer;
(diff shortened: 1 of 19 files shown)


src/main/java/software/amazon/smithy/lsp/language/RenameHandler.java
```diff
@@ -0,0 +1,260 @@
...
+            String conflictingId = getConflictingImport(sourceFile, newName);
+            if (conflictingId == null) {
+                return;
+            }
+
+            var conflictingShape = ShapeSearch.findShape(sourceFile.getParse(), conflictingId, config.model())
+                    .orElse(null);
+            if (conflictingShape == null) {
+                return;
+            }
```
yefrig, line 113: can this be improved? it looks awkward lol but maybe `Optional` is not that flexible

src/main/java/software/amazon/smithy/lsp/syntax/Syntax.java
```diff
@@ -293,14 +293,20 @@ public List<Node> elements() {
          * identifiers, so this class a single subclass {@link Ident}.
          */
         public static sealed class Str extends Node {
+            final int line;
```
yefrig, line 296: is this supposed to be line number?

yefrig (approved):
Left some small comments but otherwise LGTM.

Also tested functionality by building this PR locally and using it in Neovim 0.11.

src/main/java/software/amazon/smithy/lsp/SmithyLanguageServer.java
```diff
@@ -712,6 +724,61 @@ public CompletableFuture<List<? extends TextEdit>> formatting(DocumentFormatting
...
+        return supplyAsync(() -> handler.handle(params));
+    }
+
+    @Override
+    public CompletableFuture<Either3<Range, PrepareRenameResult, PrepareRenameDefaultBehavior>>
+    prepareRename(PrepareRenameParams params) {
+        LOGGER.finest("PrepareRename");
+        String uri = params.getTextDocument().getUri();
+        ProjectAndFile projectAndFile = state.findProjectAndFile(uri);
+        if (projectAndFile == null) {
```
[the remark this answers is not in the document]
milesziemer: Yea I agree. There's also a lot of duplicated setup code in each of the handlers, for example the `ReferencesHandler.Config` class which is shared with rename. I'm thinking I'll follow up with some changes to refactor all that

src/main/java/software/amazon/smithy/lsp/syntax/Syntax.java
milesziemer: Yes

src/main/java/software/amazon/smithy/lsp/language/ReferencesHandler.java
```diff
@@ -0,0 +1,115 @@
...
+                idlDefinitionFile = idl;
+            }
+
+            return new Config(id, shape, model, idlDefinitionFile);
+        }
+
+        private static ResponseErrorException notSupported() {
+            var error = new ResponseError();
+            error.setCode(ResponseErrorCode.RequestFailed);
+            error.setMessage("Not supported here.");
```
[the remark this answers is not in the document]
milesziemer: Updated

src/main/java/software/amazon/smithy/lsp/syntax/Syntax.java
milesziemer: Made this lineNumber

PR #222, opened 2025-04-16
Cleanup build script

milesziemer:
Our build script was a mess and it has been bugging me for a while, so I'm
finally cleaning it up. I mostly reorganized things, and removed unnecessary
configuration, but there are some minor changes to note:
- Got rid of the git-version plugin, which shouldn't be necessary since we
  use the VERSION file to manage the project version.
- Got rid of the manual checkstyle dependency, since what we were setting
   it to was already the default.
- Changed test logging to only log failed and skipped, instead of passed and
   stdout/err, which should reduce some noise.
- Fixed the dependency declaration for smithy libraries, which weren't using
   string interpolation, so were literally "[smithyVersion, 2.0[". Either that was
   a thing you could do in groovy, or it's been busted since 2023. Yikes.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,7 @@ println("Smithy Language Server version: '${version}'")
 
 val stagingDirectory = layout.buildDirectory.dir("staging")
 val releaseChangelogFile = layout.buildDirectory.file("resources/RELEASE_CHANGELOG.md").get()
+val smithyVersion: String by project
 
 plugins {
     java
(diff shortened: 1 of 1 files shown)


joewyz (approved):

build.gradle.kts
```diff
@@ -19,228 +19,216 @@ import org.jreleaser.model.Active
...
-
-println("Smithy Language Server version: '${libraryVersion}'")
-
-val stagingDirectory = rootProject.layout.buildDirectory.dir("staging")
+dependencies {
+    implementation("org.eclipse.lsp4j:org.eclipse.lsp4j:0.23.1")
+    implementation("software.amazon.smithy:smithy-build:[$smithyVersion, 2.0[")
+    implementation("software.amazon.smithy:smithy-cli:[$smithyVersion, 2.0[")
+    implementation("software.amazon.smithy:smithy-model:[$smithyVersion, 2.0[")
+    implementation("software.amazon.smithy:smithy-syntax:[$smithyVersion, 2.0[")
```
haydenbaker, line 57: You don't really need to set the version if you're using a range, just `[1.0, 2.0]`

PR #250, opened 2026-04-18
Make capability registration resilient to unresponsive clients

yasmewad:
Some LSP clients (such as Kiro CLI) don't respond to `client/registerCapability` or `client/unregisterCapability`
requests, which blocks lsp4j's message loop and hangs the server. This is especially problematic
because the unregister→register chain runs on every file change, workspace folder change, and
build file save (since clients don't de-duplicate file watchers).

Per the LSP 3.17 spec, dynamic registration is optional — *"Not all clients need to support
dynamic capability registration"* — and `didChangeWatchedFiles` has no static fallback
(*"the current protocol doesn't support static configuration for file changes from the server
side"*). The server must degrade gracefully when registration isn't available.

### Changes

- **Best-effort capability registration**: Added `registerBestEffort()` and
  `unregisterBestEffort()` helper methods that wrap all 6 `registerCapability`/
  `unregisterCapability` call sites with `.exceptionally()` handlers. Each call site
  logs a distinct warning describing which registration failed.

- **Fixed unregister→register chains**: Changed `thenRun` to `thenCompose` in the 3
  places that chain unregister then register (`didChangeWatchedFiles`,
  `didChangeWorkspaceFolders`, `didSave`). `thenRun` silently skips the re-registration
  if the upstream future completes exceptionally; `thenCompose` properly composes the
  two async operations so re-registration always proceeds.

- **Removed unused version.properties**: The `createProperties` Gradle task and startup
  banner in `connect()` were dead code. Removed along with their now-unused imports.

- **Made `StubClient` extensible**: Removed `final` modifier to allow test-specific
  subclasses that simulate client failures.

### Testing

- `initializedHandlesRegisterCapabilityFailure` — verifies the server continues when
  `registerCapability` returns a failed future during startup
- `unregisterFailureDoesNotBlockReRegistration` — verifies the unregister→register
  chain in `didChangeWatchedFiles` still works when `unregisterCapability` fails
- After using the binary generated, it works well with agentic coding tools like Kiro CLI

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -96,17 +96,6 @@ tasks {
         archiveClassifier = "javadoc"
     }
 
-    val createProperties by register("createProperties") {
-        dependsOn(processResources)
-
-        doLast {
-            val file = layout.buildDirectory.file("resources/main/version.properties").get().asFile
-            val properties = Properties()
-            properties["version"] = version.toString()
-            properties.store(file.writer(), null)
-        }
-    }
-
     // Generate a changelog that only includes the changes for the latest version
     // which JReleaser will add to the release notes of the GitHub release.
     val createReleaseChangelog by register("createReleaseChangelog") {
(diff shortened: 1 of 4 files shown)


yasmewad:
Closing as I realized the existing binary pattern works with Kiro CLI. It was just that I was missing some setup instructions in correct order.

PR #261, opened 2026-07-09
Add support to run DiffEvaluators on save

a-morales:
Note: I did use Claude to develop this feature, but I did several rounds of manual code review and cleanup before submitting this PR

This adds a new feature to be able to run [Smithy's DiffEvaluator](https://github.com/smithy-lang/smithy/tree/main/smithy-diff) on the current project being worked on against a configured baseline model. I currently work on managing a registry defined in Smithy and we use DiffEvaluators for detecting breaking changes. Once configured when a user saves any changes the DiffEvaluators will run and any errors will be reported back through LSP diagnostics. 

I added two ways to load a baseline (`BaselineProvider`). The first one is `MavenBaselineProvider` which works through Maven and takes in a standard dependency coordinate to be able to download the provider. Since Smithy's Maven Resolver doesn't support having `LATEST` in the coordinate it would be pretty tedious/fragile to update that continuously in my registry. I do have a HTTP service that serves my model as JSON so I also added a `UrlBaselineProvider`. Once a Baseline is loaded and a file is saved, DiffEvaluators are discovered and then individually run to isolate them and prevent a bad Evaluator from bringing down the others. The events we get back then get processed to be able to anchor them to an opened file or the base `.smithy-project.json` file. This is necessary since shape removals may belong to a file that was removed. 
CleanShot 2026-07-08 at 15 58 29


src/main/java/software/amazon/smithy/lsp/diff/MavenBaselineProvider.java
```diff
@@ -0,0 +1,195 @@
...
+import software.amazon.smithy.cli.dependencies.MavenDependencyResolver;
+import software.amazon.smithy.cli.dependencies.ResolvedArtifact;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.loader.ModelAssembler;
+import software.amazon.smithy.model.loader.ModelDiscovery;
+import software.amazon.smithy.model.validation.ValidatedResult;
+
+/**
+ * Baseline provider that resolves a Maven coordinate (e.g.
+ * {@code com.disneystreaming.api.registry:registry-snapshot:1.2.3}) with the same
```
lewisjkl: Remove reference to disneystreaming

src/main/java/software/amazon/smithy/lsp/diff/MavenBaselineProvider.java
```diff
@@ -0,0 +1,195 @@
...
+ * {@link DependencyResolver} machinery the language server already uses, then assembles the
+ * baseline model from the resolved jar(s).
+ *
+ * <p>Assembly mirrors {@code api-registry-cli}'s {@code ModelLoader.withoutCurrentClassloader}:
```
lewisjkl: Also this reference to api-registry

lewisjkl (commented):
Will read through the rest, but just quickly noticed these

